PR #2, opened 2021-01-15
add org:display and org:list

mshanemc:
### What does this PR do?
moves the org:display and org:list commands from toolbelt [serious refactor]
@W-8668310@
@W-8668315@

https://github.com/forcedotcom/cli/issues/482
https://github.com/forcedotcom/cli/issues/666
https://github.com/forcedotcom/cli/issues/422


test/commands/force/org/display.test.ts
```diff
@@ -0,0 +1,244 @@
...
+  });
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async function prepareStubs(AuthInfoModifications: any = {}) {
+    authInfoStub = stubInterface<AuthInfo>($$.SANDBOX, {
+      getFields: () => ({
+        ...baseAuthInfo,
+        ...AuthInfoModifications,
+      }),
+      // getConnectionOptions: () => ({ accessToken: baseAuthInfo.accessToken }),
```
peternhale, line 68: Remove if not needed

test/commands/force/org/display.test.ts
```diff
@@ -0,0 +1,244 @@
...
+    .stdout()
+    .command(['force:org:display', '--targetusername', 'scratch@test.com', '--json'])
+    .it('queries server for scratch org info', (ctx) => {
+      // console.log(ctx.stdout);
+      const result = JSON.parse(ctx.stdout).result;
+      expect(result).to.not.be.undefined;
+      expect(result.status).to.equal('Active');
+    });
+
+  it('gets non-scratch org connectedStatus');
```
peternhale, line 239: Still being written?

peternhale (changes requested):

src/commands/force/org/list.ts
```diff
@@ -0,0 +1,186 @@
...
+      nonScratchOrgs: sortBy(metaConfigs.nonScratchOrgs, this.sortFunction),
+      scratchOrgs: sortBy(metaConfigs.scratchOrgs, this.sortFunction),
+      expiredScratchOrgs: metaConfigs.scratchOrgs.filter((org) => !identifyActiveOrgByStatus(org)),
+    };
+
+    if (this.flags.clean && groupedSortedOrgs.expiredScratchOrgs.length > 0) {
+      await this.cleanScratchOrgs(groupedSortedOrgs.expiredScratchOrgs, !this.flags.noprompt);
+    }
+
+    if (groupedSortedOrgs.expiredScratchOrgs.length > 10 && !this.flags.clean) {
```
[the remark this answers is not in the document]
mshanemc: That's how it currently works in toolbelt.

My opinion: a few orgs doesn't hurt, but when there's a ton of expired scratch orgs, that's bigger lists of orgs to process, sometimes query from the server, etc.  I think we're balancing performance improvements (tell them when it's becoming significant) vs. hassling them about a single expired org.

src/shared/orgListUtil.ts
```diff
@@ -0,0 +1,275 @@
...
+      const conn = devHubOrg.getConnection();
+      const data = await conn
+        .sobject('ScratchOrgInfo')
+        .find<ExtendedScratchOrgInfo>({ ScratchOrg: { $in: orgIdsToQuery } }, fields);
+      return data.map((org) => ({
+        ...org,
+        devHubOrgId: devHubOrg.getOrgId(),
+      }));
+    } catch (err) {
+      return [];
```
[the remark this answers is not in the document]
mshanemc: yes

src/commands/force/org/list.ts
```diff
@@ -0,0 +1,186 @@
...
+      description: messages.getMessage('all'),
+    }),
+    clean: flags.boolean({
+      description: messages.getMessage('clean'),
+    }),
+    noprompt: flags.boolean({
+      char: 'p',
+      description: messages.getMessage('noPrompt'),
+    }),
+    skipconnectionstatus: flags.boolean({
```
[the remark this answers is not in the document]
mshanemc: no, but your comment made me think of something else.  It's doesn't do anything without `--clean` so I'm making it addings `dependsOn`.

test/commands/force/org/display.test.ts
mshanemc: We left pending test in toolbelt.  I'd prefer to keep around for ideas of "things we might want to add later when there's time"  

I made the list of test idea, then got coverage up to a good level, and wanted to get some ideas on this given sprint deadlines + QA.

Do we want these to not exist at all?

peternhale: Just curious, I am ok with leaving pending tests in place.

src/shared/orgListUtil.ts
```diff
@@ -0,0 +1,275 @@
...
+  ): Promise<ExtendedAuthFields[]> {
+    // Reduce the information to key value pairs with signupUsername as key
+    const contentMap = updatedContents.reduce((map, scratchOrgInfo) => {
+      if (scratchOrgInfo) {
+        map[scratchOrgInfo.SignupUsername] = scratchOrgInfo;
+      }
+      return map;
+    }, {});
+
+    for (const scratchOrgInfo of orgs) {
```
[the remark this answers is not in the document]
mshanemc: why?

src/commands/force/org/list.ts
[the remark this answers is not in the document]
peternhale: 👍

src/shared/orgListUtil.ts
```diff
@@ -0,0 +1,275 @@
...
+          if (flags.skipconnectionstatus) {
+            fields.connectedStatus = fields.connectedStatus ?? 'Unknown';
+          } else {
+            fields.connectedStatus = await this.determineConnectedStatusForNonScratchOrg(fields.username);
+          }
+          return fields;
+        })
+      ),
+
+      this.processScratchOrgs(orgs.scratchOrgs),
```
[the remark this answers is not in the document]
mshanemc: Every method but the logger stuff is.  I'm switching them all.

src/shared/orgListUtil.ts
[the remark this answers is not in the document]
peternhale: Supports more of a functional style. Personal choice, indeed.

src/shared/orgListUtil.ts
```diff
@@ -0,0 +1,275 @@
...
+      return [];
+    }
+  }
+
+  public static async reduceScratchOrgInfo(
+    updatedContents: Array<Record<ExtendedScratchOrgInfo>>,
+    orgs: ExtendedAuthFields[]
+  ): Promise<ExtendedAuthFields[]> {
+    // Reduce the information to key value pairs with signupUsername as key
+    const contentMap = updatedContents.reduce((map, scratchOrgInfo) => {
```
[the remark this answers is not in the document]
mshanemc: It's definitely an object, not a username.  What's the clarity issue it's causing?

src/shared/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Aliases } from '@salesforce/core';
+
+export const getAliasByUsername = async (username: string): Promise<string> => {
+  const alias = await Aliases.create(Aliases.getDefaultOptions());
+  const aliasContent = alias.getContents().orgs;
```
amphro, line 11: Isn't this method the same as `https://forcedotcom.github.io/sfdx-core/classes/aliases.html#getkeysbyvalue`? I guess in this case you only want one, but an interesting question is what should happen if there is more than one alias? Currently, it only displays the first. Maybe it should display all of them though.

tmp/root/cli_reference.ditamap
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
amphro, line 1: please git ignore `tmp/`

amphro (changes requested):

src/shared/orgListUtil.ts
```diff
@@ -0,0 +1,275 @@
...
+    return {
+      nonScratchOrgs,
+      scratchOrgs,
+    };
+  }
+
+  public static async processScratchOrgs(scratchOrgs: ExtendedAuthFields[]): Promise<ExtendedAuthFields[]> {
+    // organize by DevHub to reduce queries
+    const orgIdsGroupedByDevHub: Dictionary<string[]> = {};
+    scratchOrgs.forEach((fields) => {
```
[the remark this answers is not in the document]
mshanemc: say more about what you have in mind?  This is a kinda like groupby but instead of grouping the whole object, grab one property and do a small transform (18=>15 character IDs)

src/shared/orgListUtil.ts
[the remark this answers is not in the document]
mshanemc: That's what I've done, too.  Sometime things start off without async and then have to switch over and it's less work.  

That said, this file has some cut/paste from toolbelt so I've got both flavors :)

src/shared/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (aliasContent) {
+    for (const aliasedName of Object.keys(aliasContent)) {
+      if (aliasContent[aliasedName] === username) return aliasedName;
+    }
+  }
+  return undefined;
+};
+
+export const camelCaseToTitleCase = (text: string): string => {
```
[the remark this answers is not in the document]
mshanemc: added one per our poll.

package.json
```diff
@@ -73,12 +75,19 @@
       "@salesforce/plugin-command-reference"
     ],
     "topics": {
-      "hello": {
-        "description": "Commands to say hello."
+      "force": {
+        "description": "tools for the Salesforce developer",
```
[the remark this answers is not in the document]
mshanemc: will do...but how does that work?

src/shared/utils.ts
mshanemc: I don't see that method, either in core OR in your link.

tmp/root/cli_reference.ditamap
mshanemc: also added a user story to fix our template repos' gitignore

src/shared/orgListUtil.ts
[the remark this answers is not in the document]
peternhale: Sorry for not being more clear. Thinking of a lambda that accomplishes the same thing as this logic, produce a dictionary by devhub username and its associated list of scratch org ids. JS reduce does not neatly handle this as well as Java Stream, so the result is not much different, other than eliminating conditionals and removes initialization of the result var.

My example for this
```
    const orgIdsGroupedByDevHub = scratchOrgs
      .filter((fields) => fields.devHubUsername)
      .reduce((accumulator: Dictionary<string[]>, fields) => {
        const orgs = accumulator[fields.devHubUsername] ?? [];
        orgs.push(sfdc.trimTo15(fields.orgId));
        accumulator[fields.devHubUsername] = orgs;
        return accumulator;
      }, {});
```

src/shared/utils.ts
mshanemc: As far as the multi-alias issue...this utils method could bring back multiple, but the list/display commands that use this don't currently support that and switch the property from string to string[] would probably be hard on the users relying on json output.

Maybe we should prevent multiple aliases from being created for the same username.

src/shared/orgListUtil.ts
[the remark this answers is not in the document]
mshanemc: I changed it to something more along those lines.

src/shared/orgListUtil.ts
```diff
@@ -76,16 +76,12 @@ export class OrgListUtil {
...
-        if (!orgIdsGroupedByDevHub[fields.devHubUsername]) {
-          orgIdsGroupedByDevHub[fields.devHubUsername] = [];
-        }
-        orgIdsGroupedByDevHub[fields.devHubUsername].push(sfdc.trimTo15(fields.orgId));
-      }
-    });
+    const orgIdsGroupedByDevHub = scratchOrgs
+      .filter((fields) => fields.devHubUsername)
+      .reduce((accum: Dictionary<string[]>, fields) => {
+        accum[fields.devHubUsername] = [...(accum[fields.devHubUsername] ?? []), sfdc.trimTo15(fields.orgId)];
```
peternhale, line 82: Nice!

package.json
```diff
@@ -73,12 +75,18 @@
       "@salesforce/plugin-command-reference"
     ],
     "topics": {
-      "hello": {
-        "description": "Commands to say hello."
+      "force": {
+        "external": true,
```
amphro, line 79: I think you asked what it does right? It is just for https://github.com/forcedotcom/plugin-command-reference 

We could define the force topic description at every plugin that defines it, but then what if we need to update it? Which one wins? Instead, this says "I am adding commands to an existing topic" and we define the description in `salesforcedx`. This property just tells the generator not to error out because there isn't a topic description.

src/shared/utils.ts
amphro: I don't know. I can see multiple aliases on the same user being useful and since we already support that I wouldn't want to change it at this point.

This is how it currently works, so leave it. Just thought I would bring it up.

amphro (approved):

src/shared/utils.ts
mshanemc: What about the phantom `getkeysbyvalue` method?  Any idea where that is?

peternhale: @mshanemc it exists on Alias instance getKeysByValue
Screen Shot 2021-01-20 at 1 26 54 PM

mshanemc: Aliases extends ConfigGroup extends ConfigFile extends BaseConfigStore.  Inheritance...

peternhale (approved):

PR #7, opened 2021-01-27
Sm/org open

mshanemc:
### What does this PR do?
Open source org open, 50% rewrite

- use newer flag tricks on `--path`
- reuse myDomainResolver from core
- typed return object
- new isInternal from latest core
- unit tests turned out really nice

@W-8775373@


src/shared/utils.ts
```diff
@@ -18,3 +20,7 @@ export const camelCaseToTitleCase = (text: string): string => {
     .replace(/([A-Z][a-z]+)/g, ' $1')
     .trim();
 };
+
+export const openUrl = async (url: string): Promise<ChildProcess> => {
+  return open(url, { wait: false });
```
peternhale, line 25: wait defaults to false

test/commands/force/org/open.test.ts
```diff
@@ -0,0 +1,176 @@
...
+  expect(response).to.have.property('orgId').equal(orgId);
+  return true;
+};
+
+describe('open commands', () => {
+  const spies = new Map();
+  afterEach(() => spies.clear());
+
+  beforeEach(async function () {
+    $$.SANDBOX.restore();
```
peternhale, line 35: Why not restore in `afterEach`?

peternhale (changes requested):

src/commands/force/org/open.ts
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  public async run(): Promise<OrgOpenOutput> {
+    const frontDoorUrl = await this.buildFrontdoorUrl();
+    const url = `${frontDoorUrl}&retURL=${this.flags.path as string}`;
+    const orgId = this.org.getOrgId();
+    const username = this.org.getUsername();
+    const output = { orgId, url, username };
+
+    if (new Env().getString('SFDX_CONTAINER_MODE') ? true : false) {
```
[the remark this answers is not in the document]
amphro: +1

src/commands/force/org/open.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (sfdc.isInternalUrl(domain) || timeout.seconds === 0) {
+      return;
+    }
+
+    const resolver = await MyDomainResolver.create({
+      url: new URL(domain),
+      timeout,
+      frequency: new Duration(1, Duration.Unit.SECONDS),
+    });
+    this.ux.startSpinner(messages.getMessage('domainWaiting'));
```
amphro, line 82: Do we need to stop it?

test/commands/force/org/list.test.ts
```diff
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
 /*
```
amphro, line 7: no need for two

amphro (approved):
Other than addressing some of Pete's comments, looks good.

test/commands/force/org/open.test.ts
mshanemc: 🤷🏻 sometimes there's an `afterEach`, but there's almost always a beforeEach.  So it's mostly consistency (same place each time so it gets done).

src/commands/force/org/open.ts
mshanemc: It happens automatically when the command exits, right?  I see a lot of empty spinner-stopping code in the CLI and have been wondering why we do that.

Normally I only stop it if I want to say something specific in the message instead of the default "done".

Or because you think the thrown message before done makes people feel like it shouldn't say "done"?

test/commands/force/org/list.test.ts
mshanemc: whoa

test/commands/force/org/open.test.ts
peternhale: Consistency is a wonderful thing, but IMHO, teardown tasks belong in teardown methods.

src/commands/force/org/open.ts
peternhale: I ran into a weird situation in a fancy test where not stopping the spinner caused stdout not to be collected, causing a false test failure. Not sure why.

mshanemc: weird.  I had a test for stdout in non-json mode, and just added one for stderr.  Both are fine without adding the extra stopspinner code.

peternhale (approved):

package.json
```diff
@@ -7,7 +7,9 @@
   "dependencies": {
     "@oclif/config": "^1",
     "@salesforce/command": "^3.0.3",
-    "@salesforce/core": "^2.16.3",
+    "@salesforce/core": "^2.17.0",
```
peternhale, line 10: Pulled in 2.17.0 - changes needed to isInternalUrl, etc.
1.4.0 of kit was bumped "automagically"

package.json
```diff
@@ -90,20 +92,20 @@
   "scripts": {
     "build": "sf-build",
     "clean": "sf-clean",
-    "clean-all": "-clean all",
+    "clean-all": "sf-clean all",
```
peternhale, line 95: Something in prepare script is stripping `sf` from clean-all

PR #27, opened 2021-02-09
fix: no json connectedStatus on skipconnection

mshanemc:
### What does this PR do?
toolbelt: `--json --skipconnectionstatus` would have no property for connectedStatus on non-scratch orgs
plugin: set the connectedStatus to 'Unknown'

This change fixes that.

We also never update isDevHub, and a viable customer scenario is using a prod org for CLI work (ex w/ sandboxes) and then turning on DevHub to get scratch orgs.  Now `org:list` re-checks on the server. https://github.com/forcedotcom/cli/issues/482 

h/t @SCWells72 for all the help

@W-8871284@
@W-8881661@ https://github.com/forcedotcom/cli/issues/482

### DIFF
--- a/src/shared/orgListUtil.ts
+++ b/src/shared/orgListUtil.ts
@@ -55,7 +55,7 @@ export class OrgListUtil {
           if (!flags.skipconnectionstatus) {
             // skip completely if we're skipping the connection
             fields.connectedStatus = await OrgListUtil.determineConnectedStatusForNonScratchOrg(fields.username);
-            if (!fields.isDevHub) {
+            if (!fields.isDevHub && fields.connectedStatus === 'Connected') {
               // activating DevHub setting is irreversible so don't waste time checking any org we already know is a hub
               fields.isDevHub = await OrgListUtil.checkNonScratchOrgIsDevHub(fields.username);
             }
--- a/src/shared/orgListUtil.ts
+++ b/src/shared/orgListUtil.ts
@@ -52,9 +52,7 @@ export class OrgListUtil {
     const [nonScratchOrgs, scratchOrgs] = await Promise.all([
       Promise.all(
         orgs.nonScratchOrgs.map(async (fields) => {
-          if (flags.skipconnectionstatus) {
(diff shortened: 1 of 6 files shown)


src/shared/orgListUtil.ts
```diff
@@ -52,9 +52,7 @@ export class OrgListUtil {
     const [nonScratchOrgs, scratchOrgs] = await Promise.all([
       Promise.all(
         orgs.nonScratchOrgs.map(async (fields) => {
-          if (flags.skipconnectionstatus) {
-            fields.connectedStatus = fields.connectedStatus ?? 'Unknown';
-          } else {
+          if (!flags.skipconnectionstatus) {
```
peternhale, line 55: @mshanemc is there a test to cover this?

mshanemc: https://github.com/salesforcecli/plugin-org/blob/c53fba1f9acd6ceb5318bf73727ae851162fee54/test/shared/orgListUtil.test.ts#L112

the command is just passing that flag to orgListUtil so the test is at that level instead of on the command.

amphro: Another way to put it, is there a way this bug could have been caught with a test?

mshanemc: that test now asserts that `--skipconnectionstatus` produces nonScratchOrgs with no `connectedStatus` in the json

peternhale: @mshanemc Does the test below represent verification of the correct behavior when `flags.skipconnectionstatus` is false?

https://github.com/salesforcecli/plugin-org/blob/de91ff54a7db5d78460293e8889cbf9b703f8514/test/shared/orgListUtil.test.ts#L102

mshanemc: when migrating this command, I tried to keep the tests on orgListUtil as close to the original as possible.

I don't like the original either, so I'm gonna fix them.

mshanemc: better?

peternhale (approved):
Thank you!

PR #28, opened 2021-02-18
fix: refresh before reading files

mshanemc:
### What does this PR do?
refresh auth before reading the files to prevent display of stale accessTokens

@W-8886636@

### DIFF
--- a/src/commands/force/org/display.ts
+++ b/src/commands/force/org/display.ts
@@ -27,6 +27,13 @@ export class OrgDisplayCommand extends SfdxCommand {
   };
 
   public async run(): Promise<OrgDisplayReturn> {
+    try {
+      // the auth file might have a stale access token.  We want to refresh it before getting the fields
+      await this.org.refreshAuth();
+    } catch (error) {
+      // even if this fails, we want to display the information we can read from the auth file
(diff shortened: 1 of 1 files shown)


WillieRuemmele (approved):

src/commands/force/org/display.ts
```diff
@@ -27,6 +27,13 @@ export class OrgDisplayCommand extends SfdxCommand {
   };
 
   public async run(): Promise<OrgDisplayReturn> {
+    try {
+      // the auth file might have a stale access token.  We want to refresh it before getting the fields
+      await this.org.refreshAuth();
```
peternhale, line 32: What if the access token is not stale? Any way to check before forcing the auth?

mshanemc: That's already built in with refreshAuth().  It's definitely only updating them if they went stale...I assume it's trying to use them, catching expirations, and using the refreshToken (where available) or re-jwt-ing.

amphro: Correct. It tries to use it and if the server says it is invalid or expired, jsforce will kick off the refresh flow that we defined.

amphro (approved):

peternhale (approved):

PR #54, opened 2021-03-04
chore: add nut tests for plugin-org

peternhale:
### What does this PR do?
Add new nut tests for plugin-org
@W-8856585@

### DIFF
--- a/src/shared/orgHighlighter.ts
+++ b/src/shared/orgHighlighter.ts
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import chalk from 'chalk';
+import chalk = require('chalk');
 import { ExtendedAuthFields } from './orgTypes';
 
 /* eslint-disable @typescript-eslint/no-unsafe-return */
--- a/test/nut/commands/force/org/org.nut.ts
+++ b/test/nut/commands/force/org/org.nut.ts
@@ -36,7 +36,7 @@ describe('Org Command NUT', function async() {
       defaultUsername = getString(session.setup[0], 'result.username');
       defaultUserOrgId = getString(session.setup[0], 'result.orgId');
       aliasedUsername = getString(session.setup[1], 'result.username');
-      aliasUserOrgId = getString(session.setup[2], 'result.orgId');
+      aliasUserOrgId = getString(session.setup[1], 'result.orgId');
     }
   });
 
@@ -129,7 +129,7 @@ describe('Org Command NUT', function async() {
       expect(aliasUserLine).to.include('anAlias');
     });
   });
-  describe.skip('Org Display', () => {
+  describe('Org Display', () => {
     it('should display org information for default username', async () => {
       const result = execCmd('force:org:display --json');
       expect(result.shellOutput.code).to.be.equal(
@@ -181,7 +181,7 @@ describe('Org Command NUT', function async() {
   });
   describe('Org Open', () => {
     it('should produce the URL for an org in json', async () => {
-      const result = execCmd('force:org:open -u dh@salesforce.com --urlonly --json');
+      const result = execCmd(`force:org:open -u ${defaultUsername} --urlonly --json`);
       expect(result.shellOutput.code).to.be.equal(
         0,
         concatString(result.shellOutput.stdout, result.shellOutput.stderr)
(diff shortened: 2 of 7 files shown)


src/shared/orgHighlighter.ts
```diff
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import chalk from 'chalk';
+import chalk = require('chalk');
```
mshanemc, line 7: <head scratch> some kind of types/TS setup problem?

test/nut/commands/force/org/org.nut.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+  });
+
+  after(async () => {
+    await session?.clean();
+  });
+
+  describe('List Orgs', () => {
+    it('should list all orgs', () => {
+      const result = execCmd('force:org:list --json');
```
mshanemc, line 50: Totally works.  FYI: Mike added TS generics to execCmd to reduce the amount of casting that has to be done, which pairs nicely when commands export their return type.

example: https://github.com/salesforcecli/plugin-user/blob/40eb04a31d446cf480c524d5ab245621d5dec987/test/allCommands.nut.ts#L36

test/nut/commands/force/org/org.nut.ts
```diff
@@ -0,0 +1,214 @@
...
+          connectedStatus: 'Connected',
+        },
+        JSON.stringify(nonScratchOrgs)
+      );
+    });
+    it('should list orgs - skipconnectionstatus', () => {
+      const result = execCmd('force:org:list --skipconnectionstatus --json');
+      expect(result.shellOutput.code).to.be.equal(
+        0,
+        concatString(result.shellOutput.stdout, result.shellOutput.stderr)
```
mshanemc, line 87: This does produce nicer messages from mocha on fails.  Does this belong in testkit so Pete doesn't have to boilerplate it?  @shetzel 

https://github.com/salesforcecli/cli-plugins-testkit/blob/a018e3f43572df1c9c927c341eea1ed393b0ef6a/src/execCmd.ts#L81

mshanemc (approved):
approved with comments.  The only one I have strong feelings about it the weird import/require change

test/nut/commands/force/org/org.nut.ts
```diff
@@ -0,0 +1,214 @@
...
+import { TestSession } from '@salesforce/cli-plugins-testkit';
+import { execCmd } from '@salesforce/cli-plugins-testkit';
+import { Duration } from '@salesforce/kit';
+import { asDictionary, AnyJson, get, getString, isArray } from '@salesforce/ts-types';
+
+const concatString = (a: string, b: string): string => a + b;
+
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+
+describe('Org Command NUT', function async() {
```
WillieRuemmele, line 20: could we make this an arrow function for consistency?

test/shared/orgHighlighter.test.ts
```diff
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { expect } from '@salesforce/command/lib/test';
-import chalk from 'chalk';
+import chalk = require('chalk');
```
WillieRuemmele, line 8: same comment as Shane's above

test/nut/commands/force/org/org.nut.ts
mshanemc: Definitely not here because Pete's using the mocha `this` which the arrow loses.  
```
this.timeout(Duration.minutes(10).milliseconds);
```

Mocha's opinion on the subject: https://mochajs.org/#arrow-functions

peternhale: @WillieRuemmele, I am increasing timeout here and I need access to mocha's `this`. The difficulty I ran into was testing against local build. The setup takes longer, so overriding default applied in test:nuts.

I can remove the timeout and switch to arrow function.

test/nut/commands/force/org/org.nut.ts
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+import * as os from 'os';
+import * as querystring from 'querystring';
+import { expect } from '@salesforce/command/lib/test';
+import { TestSession } from '@salesforce/cli-plugins-testkit';
+import { execCmd } from '@salesforce/cli-plugins-testkit';
+import { asDictionary, AnyJson, Dictionary, getString, isArray } from '@salesforce/ts-types';
+
+/* eslint-disable @typescript-eslint/no-unsafe-call */
```
shetzel, line 15: If we disable rules we must add a comment explaining why.  Also, is it really worth disabling the rule if we can just write the code in a slightly different way?  Pretty sure this is so we can do: `expect(devHubLine).to.be.ok;` which is arguably unnecessary.

shetzel (approved):
Looks good!  1 non-blocking comment.

test/nut/commands/force/org/org.nut.ts
peternhale: @shetzel I will remove it and the `ok` checks

shetzel (approved):

mshanemc (approved):

PR #56, opened 2021-03-11
fix: clean up help and add examples back to force:org:list

jshackell-sfdc:
### What does this PR do?

Cleans up help and adds examples back to force:org:list

@W-8995248@

### DIFF
--- a/messages/open.json
+++ b/messages/open.json
@@ -1,5 +1,5 @@
 {
-  "description": "open your default scratch org, or another specified org\nTo open a specific page, specify the portion of the URL after \"yourInstance.salesforce.com/\" as --path.\nFor example, specify \"--path lightning\" to open Lightning Experience, or specify \"--path /apex/YourPage\" to open a Visualforce page.\nTo generate a URL but not launch it to your browser, specify --urlonly.",
+  "description": "open your default scratch org, or another specified org\nTo open a specific page, specify the portion of the URL after \"yourInstance.salesforce.com/\" as --path.\nFor example, specify \"--path lightning\" to open Lightning Experience, or specify \"--path /apex/YourPage\" to open a Visualforce page.\nTo generate a URL but not launch it in your browser, specify --urlonly.",
   "examples": [
     "sfdx force:org:open",
     "sfdx force:org:open -u me@my.org",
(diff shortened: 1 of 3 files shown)


messages/open.json
```diff
@@ -1,5 +1,5 @@
 {
-  "description": "open your default scratch org, or another org that you specify,\nTo open a specific page, specify the portion of the URL after \"yourInstance.salesforce.com/\" as --path.\nFor example, specify \"--path lightning\" to open Lightning Experience, or specify \"--path /apex/YourPage\" to open a Visualforce page.\nTo generate a URL but not launch your browser, specify --urlonly",
+  "description": "open your default scratch org, or another specified org\nTo open a specific page, specify the portion of the URL after \"yourInstance.salesforce.com/\" as --path.\nFor example, specify \"--path lightning\" to open Lightning Experience, or specify \"--path /apex/YourPage\" to open a Visualforce page.\nTo generate a URL but not launch it to your browser, specify --urlonly.",
```
mshanemc, line 2: do we launch a url "**to** your browser?"  

alternatives:
launch it **in** your browser
launch it **via** your browser

mshanemc (approved):

messages/open.json
jshackell-sfdc: yep, you're right. I like "launch it IN your browser." i'll change.

src/commands/force/org/list.ts
```diff
@@ -17,7 +19,7 @@ const messages = Messages.loadMessages('@salesforce/plugin-org', 'list');
 
 export class OrgListCommand extends SfdxCommand {
   public static readonly description = messages.getMessage('description');
-
+  public static readonly examples = messages.getMessage('examples').split(EOL);
```
RodEsp, line 22: Is this for splitting up the [`description` in `messages/open.json` above](https://github.com/salesforcecli/plugin-org/pull/56/files#diff-1b7491620d62423ab4f9c0cd434d2678f216df0622da8eec99883bdeeeda4bb7R2)?

If so this won't work unless it is running on an OS that uses `\n` line endings. It might be better to just explicitly use `.split('\n')` since that's hardcoded in the `.json` file.

jshackell-sfdc: I noticed that force:org:list --help output didn't include examples, so I just copied what force:org:open does, which IS displaying the examples.  So is force:org:open also wrong?  

Full disclosure: I'm not a developer, just the tech writer, and don't usually touch real code.  But I looked at this and thought maybe I could fix it by simply copying what another command does.  But if the command I copied from is wrong, well, maybe I should just bow out...

PR #100, opened 2021-05-18
feat: support targeting of browser in "open" command

toddhalfpenny:
### What does this PR do?
This PR adds an optional `--browser` flag to the `open` command to enable users to specify which browser application to use (e.g. Firefox, Chrome). 
It does this by passing the argument through to the underlying "open" package.

An example (on mac) is;
```
$ sfdx force:org:open -u MyAlias -b "google chrome"
```

This addresses the feature request https://github.com/forcedotcom/cli/issues/1015


mshanemc:
@toddhalfpenny public contributions kinda run like this:

toddhalfpenny:
Cheers @mshanemc.
Please let me know if you need anything further from my side.

messages/open.json
```diff
@@ -1,16 +1,18 @@
 {
-  "description": "open your default scratch org, or another specified org\nTo open a specific page, specify the portion of the URL after \"yourInstance.salesforce.com/\" as --path.\nFor example, specify \"--path lightning\" to open Lightning Experience, or specify \"--path /apex/YourPage\" to open a Visualforce page.\nTo generate a URL but not launch it in your browser, specify --urlonly.",
+  "description": "open your default scratch org, or another specified org\nTo open a specific page, specify the portion of the URL after \"yourInstance.salesforce.com/\" as --path.\nFor example, specify \"--path lightning\" to open Lightning Experience, or specify \"--path /apex/YourPage\" to open a Visualforce page.\nTo generate a URL but not launch it in your browser, specify --urlonly.\nTo open in a specific browser specify the app name according to your OS as \"--browser firefox",
```
jshackell-sfdc, line 2: I suggest we add a bit more info to help the user figure out their browser app name. Here's an edit of your addition:

To open in a specific browser, specify the browser's app name according to your operating system.  For example, Chrome's app name is "google chrome" on macOS, "google-chrome" on Linux and "chrome" on Windows.  So to open an org in Chrome on macOS, specify --browser "google chrome". 

(I didn't add any escapes for the quotes, I'll let you do that :-)

messages/open.json
```diff
@@ -1,16 +1,18 @@
...
   "examples": [
     "sfdx force:org:open",
     "sfdx force:org:open -u me@my.org",
     "sfdx force:org:open -u MyTestOrg1",
-    "sfdx force:org:open -r -p lightning"
+    "sfdx force:org:open -r -p lightning",
+    "sfdx force:org:open -u me@my.org -b firefox"
   ],
+  "browser": "browser to target",
```
jshackell-sfdc, line 10: Small suggested edit:

browser in which the org opens

(okay?  I'm not thrilled with "in which" tho.  maybe "browser where the org opens"?  Your original "browser to target" seemed to be calling out for a wee more detail.  But maybe it's fine as-is.  Maybe I'm overthinking this.  I'll let you decide.)

jshackell-sfdc:
Hi @toddhalfpenny. Cool new flag :).   See my suggestions for the --help.

src/shared/utils.ts
```diff
@@ -14,6 +14,12 @@ export const getAliasByUsername = async (username: string): Promise<string> => {
   return keys?.length ? keys[0] : undefined;
 };
 
-export const openUrl = async (url: string): Promise<ChildProcess> => {
-  return open(url);
+interface openArgs {
+  app?: {
+    name: string;
+  };
+}
```
shetzel, line 21: We shouldn't need this since the open lib exports the `Options` type.

src/shared/utils.ts
```diff
@@ -14,6 +14,12 @@ export const getAliasByUsername = async (username: string): Promise<string> => {
...
+
+export const openUrl = async (url: string, args: openArgs): Promise<ChildProcess> => {
+  return open(url, args);
```
shetzel, line 24:
```suggestion
export const openUrl = async (url: string, options: open.Options): Promise<ChildProcess> => {
  return open(url, options);
}
```

shetzel (changes requested):
Small changes requested, but excellent overall.  Nice tests too!  👍

toddhalfpenny:
Thanks for all the comments and suggestions. I did a bit more digging and it appears that _open_ has [some support](https://www.npmjs.com/package/open#app) to handle x-platform browser names (at least for Chrome and Firefox). I think I'll make some adjustments to parse the inputted `browser` a bit so as to use their approach if possible (e.g. `name: open.apps.chrome`).

This should make the functionality (and doc) more user friendly and robust too. I'm thinking of adding a `createOpenOptions()` to `utils.ts`; I hope this makes sense.

toddhalfpenny:
New commit made. I have added the `utils.createOpenOptions()` function along with tests for it, as well as updating the _messages_ as suggested.

shetzel (approved):
Awesome!  👍

mshanemc:
some QA notes @toddhalfpenny 

I'm testing on macos.
when I run `./bin/run force:org:open -u hub --browser safari` I get the page opening in **both** chrome and safari?
when I run `./bin/run force:org:open -u hub --browser firefix` the command acts like everything is ok but opens nothing [you'd probably prefer to tell the user that that browser doesn't exist because it's a typo or not installed (like Edge on my machine)]

As I'm playing with this, I think I'd prefer that property to only be the known, supported options so that
1. we can tell the user what their choices are (using options in the flags)
2. if they try something else it throws an error from the flags
3. when `open` adds something else (safari?), we can add it as an option and put it in the Release Notes/help vs it just showing up stealthily.

Thoughts?

toddhalfpenny:
Thanks for the comments @mshanemc .

> when I run ./bin/run force:org:open -u hub --browser safari I get the page opening in both chrome and safari?
I had not experienced this in my tests. Very strange.

And yes, you're suggestion of limiting the supported browsers might be better. I was unable to find a way to see if the `open` command failed to work. Perhaps I'll see if I can get some info on this from its project team. In the meantime though I think restricing - as you suggest - is the best route.

jshackell-sfdc:
@toddhalfpenny , just a friendly reminder to also update the --help to reflect your changes.  I'm sure you know that, but thought I'd mention it anyway!  Right now it tells you to figure out the browser name on your own -- a discrete list of possible values will make it easier for customers.  Or at least those customers who have the standard browsers!  at-mention me when you're done updating and I can do a quick edit. Cheers!

toddhalfpenny:
Hi all, I've just created PR #211 to replace this.

It is a replica - more or less - of what was in this PR , but I was a plum and forgot to actually commit and push the last set of changes. With the amount of time and commits that have passed I thought it were best to rebase and create a new PR, instead of polluting it with lots of unrelated changes/merges etc.

Hope this is an acceptable approach, and apologies for not realising my mistake earlier and resolving sooner.

PR #124, opened 2021-06-02
fix: use most recent alias

mshanemc:
### What does this PR do?
when there are multiple aliases for the username, use the last (most recent) alias

@W-9351642@
https://github.com/forcedotcom/cli/issues/1031

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -11,7 +11,8 @@ import * as open from 'open';
 export const getAliasByUsername = async (username: string): Promise<string> => {
   const alias = await Aliases.create(Aliases.getDefaultOptions());
   const keys = alias.getKeysByValue(username);
-  return keys?.length ? keys[0] : undefined;
+  // use the most recently added alias for that username
+  return keys?.length ? keys[keys.length - 1] : undefined;
 };
 
 export const openUrl = async (url: string): Promise<ChildProcess> => {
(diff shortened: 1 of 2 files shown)


src/shared/utils.ts
```diff
@@ -11,7 +11,8 @@ import * as open from 'open';
 export const getAliasByUsername = async (username: string): Promise<string> => {
   const alias = await Aliases.create(Aliases.getDefaultOptions());
   const keys = alias.getKeysByValue(username);
-  return keys?.length ? keys[0] : undefined;
+  // use the most recently added alias for that username
+  return keys?.length ? keys[keys.length - 1] : undefined;
```
peternhale, line 15: @mshanemc is alias list associated with a user guaranteed to remain in order aliases were created?

test/shared/utils.test.ts
```diff
@@ -28,7 +28,7 @@ describe('getAliasByUsername', () => {
   });
 
   it('returns first alias for a username that has multiple aliases', async () => {
```
peternhale, line 30: @mshanemc The test name needs to change.

peternhale (changes requested):

src/shared/utils.ts
mshanemc: 1. That's how it seems to behave.
2. We're not doing any sorting of the object keys in core that I could see
https://github.com/forcedotcom/sfdx-core/blob/107070d43cfc223c7b4fb9bfa6c1ec1c10166b97/src/config/configGroup.ts#L217

maggiben (approved):

maggiben (approved):

PR #166, opened 2021-07-26
chore: using sfdx-core utill function

vamsimundra:
### What does this PR do?
It uses checkLightningDomain of sfdx-core in force:org:open command.
@W-9369342@

### DIFF
--- a/src/commands/force/org/open.ts
+++ b/src/commands/force/org/open.ts
@@ -7,7 +7,7 @@
 import { EOL } from 'os';
 
 import { flags, FlagsConfig, SfdxCommand } from '@salesforce/command';
-import { Messages, Org, checkLightningDomain, SfdxError } from '@salesforce/core';
+import { Messages, Org, SfdcUrl, SfdxError } from '@salesforce/core';
 import { Duration, Env } from '@salesforce/kit';
 import { openUrl } from '../../../shared/utils';
 
@@ -52,10 +52,11 @@ export class OrgOpenCommand extends SfdxCommand {
     // we actually need to open the org
     try {
       this.ux.startSpinner(messages.getMessage('domainWaiting'));
-      await checkLightningDomain(url);
+      const sfdcUrl = new SfdcUrl(url);
+      await sfdcUrl.checkLightningDomain();
     } catch (err) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,  @typescript-eslint/no-unsafe-call
-      if (err.message && err.message.includes('timeout')) {
+      if (err.message?.includes('timeout')) {
         const domain = `https://${/https?:\/\/([^.]*)/.exec(url)[1]}.lightning.force.com`;
         const timeout = new Duration(new Env().getNumber('SFDX_DOMAIN_RETRY', 240), Duration.Unit.SECONDS);
         this.logger.debug(`Did not find IP for ${domain} after ${timeout.seconds} seconds`);
--- a/src/commands/force/org/open.ts
+++ b/src/commands/force/org/open.ts
@@ -54,12 +54,16 @@ export class OrgOpenCommand extends SfdxCommand {
       this.ux.startSpinner(messages.getMessage('domainWaiting'));
       await checkLightningDomain(url);
     } catch (err) {
-      const domain = `https://${/https?:\/\/([^.]*)/.exec(url)[1]}.lightning.force.com`;
-      const timeout = new Duration(new Env().getNumber('SFDX_DOMAIN_RETRY', 240), Duration.Unit.SECONDS);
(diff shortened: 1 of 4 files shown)


src/commands/force/org/open.ts
```diff
@@ -51,7 +50,17 @@ export class OrgOpenCommand extends SfdxCommand {
...
     }
     // we actually need to open the org
-    await this.checkLightningDomain(url);
+    try {
+      this.ux.startSpinner(messages.getMessage('domainWaiting'));
+      await checkLightningDomain(url);
+    } catch (err) {
+      const domain = `https://${/https?:\/\/([^.]*)/.exec(url)[1]}.lightning.force.com`;
+      const timeout = new Duration(new Env().getNumber('SFDX_DOMAIN_RETRY', 240), Duration.Unit.SECONDS);
+      this.logger.debug(`Did not find IP for ${domain} after ${timeout.seconds} seconds`);
```
vamsimundra, line 59: What do you think is the best way to throw the error?
Irrespective of what error we are catching, I am throwing  "domainTimeoutError" as it was used previously. What do you think of this?

maggiben: I think the error is quite descriptive, the function will throw after the timeout+retries end so is safe to say it fits the description, the key `domainTimeoutError` seems descriptive enough to me.

src/commands/force/org/open.ts
```diff
@@ -5,11 +5,10 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { EOL } from 'os';
-import { URL } from 'url';
 
 import { flags, FlagsConfig, SfdxCommand } from '@salesforce/command';
-import { Messages, Org, MyDomainResolver, SfdxError, sfdc } from '@salesforce/core';
-import { Env, Duration } from '@salesforce/kit';
+import { Messages, Org, checkLightningDomain, SfdxError } from '@salesforce/core';
```
maggiben, line 10: Be aware that I forgot to add an export for `checkLightningDomain` here: https://github.com/forcedotcom/sfdx-core/blob/main/src/exported.ts and thus Im not sure you'll be able to load this dependency this way maybe if you do: `import checkLightningDomain from '@salesforce/core/util/checkLightningDomain'` maybe.

maggiben (approved):

test/commands/force/org/open.test.ts
```diff
@@ -7,6 +7,7 @@
 import { $$, expect, test } from '@salesforce/command/lib/test';
 import { Org, MyDomainResolver, Messages } from '@salesforce/core';
 import { stubMethod } from '@salesforce/ts-sinon';
+// import { SfdxError } from '../../../../../sfdx-core/lib/sfdxError';
```
WillieRuemmele, line 10: let's remove this comment

src/commands/force/org/open.ts
mshanemc: we're keeping the checkLightningDomain for backward-compatibility reasons. This command can use the checkLightningDomain on sfdcUrl class (validates url, etc).

https://github.com/forcedotcom/sfdx-core/pull/420/files

src/commands/force/org/open.ts
```diff
@@ -51,7 +50,21 @@ export class OrgOpenCommand extends SfdxCommand {
       return output;
     }
     // we actually need to open the org
-    await this.checkLightningDomain(url);
+    try {
+      this.ux.startSpinner(messages.getMessage('domainWaiting'));
+      await checkLightningDomain(url);
+    } catch (err) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,  @typescript-eslint/no-unsafe-call
+      if (err.message && err.message.includes('timeout')) {
```
mshanemc, line 58:
```suggestion
      if (err.message?.includes('timeout')) {
```
evaluates to undefined (falsy to the `if`) when there is no message.

src/commands/force/org/open.ts
```diff
@@ -51,7 +50,21 @@ export class OrgOpenCommand extends SfdxCommand {
...
+        const domain = `https://${/https?:\/\/([^.]*)/.exec(url)[1]}.lightning.force.com`;
+        const timeout = new Duration(new Env().getNumber('SFDX_DOMAIN_RETRY', 240), Duration.Unit.SECONDS);
```
mshanemc, line 60: where did the 240 come from?  Is that the default but it's in core?

mshanemc (approved):

mshanemc (changes requested):
needs a bump of core so the tests pass.

mshanemc (approved):

PR #186, opened 2021-08-20
feat: namespace on orgList

mshanemc:
### What does this PR do?
adds the scratch org namespace value to json output of org:list (for parity...org display already had that field) 

@W-9788406@
https://github.com/forcedotcom/cli/issues/1146

### DIFF
--- a/src/shared/orgListUtil.ts
+++ b/src/shared/orgListUtil.ts
@@ -271,6 +271,7 @@ export class OrgListUtil {
         scratchOrgInfo.status = updatedOrgInfo.Status;
         scratchOrgInfo.expirationDate = updatedOrgInfo.ExpirationDate;
         scratchOrgInfo.isExpired = updatedOrgInfo.Status === 'Deleted';
+        scratchOrgInfo.namespace = updatedOrgInfo.Namespace;
         // the old toolbelt code always said Unknown.  I'd love to get rid of it.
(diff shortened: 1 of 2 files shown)


src/shared/orgListUtil.ts
```diff
@@ -271,6 +271,7 @@ export class OrgListUtil {
         scratchOrgInfo.status = updatedOrgInfo.Status;
         scratchOrgInfo.expirationDate = updatedOrgInfo.ExpirationDate;
         scratchOrgInfo.isExpired = updatedOrgInfo.Status === 'Deleted';
+        scratchOrgInfo.namespace = updatedOrgInfo.Namespace;
```
peternhale, line 274: @mshanemc is there a test that can be modified to cover this change?

peternhale (changes requested):

src/shared/orgListUtil.ts
mshanemc: I can NUT-verify the field exists in the output and is null for the scratch orgs it creates.

The problem with namespaces is that when you change hubs, they'd fail (and some hubs might not have a NS so even a query for NS is problematic) so it's more brittle to verify an actual namespace.

peternhale (approved):

peternhale (approved):

WillieRuemmele (approved):

PR #199, opened 2021-09-27
chore: remove deprecation message

maggiben:
### What does this PR do?
Removes deprecation message
@W-9740004@


mshanemc (changes requested):
also, remove the property

mshanemc (approved):

PR #215, opened 2021-10-11
fix: warnings for json, too

mshanemc:
### What does this PR do?
moves security warning from print method to the main method so it is in warnings for `--json` like the other commands (consistency)

@W-10016839@
https://github.com/forcedotcom/cli/issues/1229

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,3 +3,4 @@
 # @forcedotcom/pdt will be requested for
 # review when someone opens a pull request.
 *       @forcedotcom/pdt
+#ECCN:Open Source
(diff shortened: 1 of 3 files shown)


src/commands/force/org/display.ts
```diff
@@ -61,14 +61,15 @@ export class OrgDisplayCommand extends SfdxCommand {
...
+    this.ux.warn(sharedMessages.getMessage('SecurityWarning'));
+
     if (!this.flags.json) {
       this.print(returnValue);
     }
     return returnValue;
   }
 
   private print(result: OrgDisplayReturn): void {
-    this.ux.warn(sharedMessages.getMessage('SecurityWarning'));
```
peternhale, line 71: @mshanemc Not issuing the message when `--json` output requested was intentional.

PR #292, opened 2022-02-14
fix: try to create local config first

maggiben:
### What does this PR do?
Tries to create local config first then falls back to global config in order to set the default username

### How to QA this:
1) if you want you can use `./bin/run` or link this plugin in sfdx by doing `sfdx plugins:link .` in the root of this folder
2) Using an org like `dreamhouse-lwc` create a new scratch org with the legacy command: `sfdx force:org:create -f config/project-scratch-def.json --setalias dreamhouse-lwc-legacy --durationdays 1 --setdefaultusername` 
3) Check out the default username with `sfdx config:list`
4) Create a new scratch org with beta command: `sfdx force:org:beta:create -f config/project-scratch-def.json --setalias dreamhouse-lwc --durationdays 1 --setdefaultusername`
5) Finally check the default username has changed with: `sfdx config:list`
@W-10675890@

### DIFF
--- a/src/commands/force/org/beta/create.ts
+++ b/src/commands/force/org/beta/create.ts
@@ -269,14 +269,15 @@ export class Create extends SfdxCommand {
       this.logger.debug('Set Alias: %s result: %s', this.flags.setalias, result);
     }
     if (this.flags.setdefaultusername) {
+      let config: Config;
       try {
-        const config = await Config.create({ isGlobal: false });
-        const result = config.set(Config.DEFAULT_USERNAME, username);
-        await config.write();
-        this.logger.debug('Set defaultUsername: %s result: %s', this.flags.setdefaultusername, result);
-      } catch (error) {
-        this.logger.debug('Set defaultUsername failed with error: %s', error);
+        config = await Config.create({ isGlobal: false });
+      } catch {
+        config = await Config.create({ isGlobal: true });
       }
+      const result = config.set(Config.DEFAULT_USERNAME, username);
+      await config.write();
+      this.logger.debug('Set defaultUsername: %s result: %s', this.flags.setdefaultusername, result);
     }
   }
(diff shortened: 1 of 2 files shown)


src/commands/force/org/beta/create.ts
```diff
@@ -269,9 +269,14 @@ export class Create extends SfdxCommand {
...
-      globalConfig.set(Config.DEFAULT_USERNAME, username);
-      const result = await globalConfig.write();
+      let config: Config;
+      try {
+        config = await Config.create({ isGlobal: false });
+      } catch {
+        config = await Config.create({ isGlobal: true });
+      }
+      const result = config.set(Config.DEFAULT_USERNAME, username);
+      await config.write();
```
shetzel, line 279: When setting the default username during org create, we only want to set it locally, not globally.  It would be a breaking change if we set it globally.  I think I would just debug log if the config doesn't update successfully and maybe add warning output.  This change needs a test as well.

shetzel (changes requested):

src/commands/force/org/beta/create.ts
shetzel: I'm sorry!  I was wrong about setting the global default.  Since `org:create` doesn't require a project you can run the command from anywhere and if you do that, it'll fail to set the local default but set it globally.  This is the expected and desired behavior.

shetzel (approved):

shetzel:
### Testing Results
✅ create org in project without alias as default sets the username in local config
❌ create org in project with alias as default sets the alias in local config - (does not match legacy command)
✅ create org outside of project without alias as default sets the username in global config
❌ create org outside of project with alias as default sets the alias in global config - (does not match legacy command)
✅ create org in project NOT as default does not set the username in local config
✅ create org outside of project NOT as default does not set the username in global config

shetzel:
With the latest updates, all tests pass.  Merging...

PR #316, opened 2022-04-21
test: for open and status commands

maggiben:
### What does this PR do?
Adds two NUTs for commands `open` and `status`

@W-10100779@


test/nut/commands/force/org/status.nut.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+
+describe('test sandbox status command', () => {
+  const env = new Env();
+  let username: string;
+
+  before(async () => {
+    username = ensureString(env.getString('TESTKIT_HUB_USERNAME'));
+    sandboxNameIn = ensureString(env.getString('TESTKIT_SANDBOX_NAME'));
```
mshanemc, line 53: suggestion: instead of specify the sbox name, get status for the first result of the query.  Then, as long as there is at least 1 sandbox regardless of its name, you're good.

mshanemc (commented):
besides my no-new-env suggestion,

also consider whether you want this `status` nut to run every time or only when the other sbox nuts run (see Willie's logic/naming conventions in the config.yml)

test/nut/commands/force/org/status.nut.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+describe('test sandbox status command', () => {
+  const env = new Env();
+  let username: string;
+
+  before(async () => {
+    username = ensureString(env.getString('TESTKIT_HUB_USERNAME'));
+    const queryStr =
+      "SELECT Id, Status, SandboxName, SandboxInfoId, LicenseType, CreatedDate, CopyProgress, SandboxOrganization, SourceId, Description, EndDate FROM SandboxProcess WHERE Status != 'E' and Status != 'D' ORDER BY CreatedDate DESC LIMIT 1";
```
WillieRuemmele, line 54: NIT: do we really need all of these fields? I only see `SandboxName` being used

test/nut/commands/force/org/status.nut.ts
```diff
@@ -0,0 +1,139 @@
...
+    const connection = await Connection.create({
+      authInfo: await AuthInfo.create({ username }),
+    });
+    const queryResult = (await connection.tooling.query(queryStr)) as { records: SandboxProcessObject[] };
+    expect(queryResult?.records?.length).to.equal(1);
+    sandboxNameIn = queryResult?.records[0]?.SandboxName;
```
WillieRuemmele, line 60: NIT: could this be refactored as `sandboxName`? the `In` is throwing me off, I'm not sure what `sandboxNameIn` means. Is it the sandbox names in na40?

test/nut/commands/force/org/status.nut.ts
```diff
@@ -0,0 +1,139 @@
...
+  it('sandbox status command sets setdefaultusername', async () => {
+    const orgStatusResult = execCmd<SandboxProcessObject>(
+      `force:org:status --sandboxname ${sandboxNameIn} -u ${username} -s --json`,
+      {
+        ensureExitCode: 0,
+      }
+    ).jsonOutput.result;
+    expect(orgStatusResult).to.be.ok;
+
+    const config = JSON.parse(
```
WillieRuemmele, line 107: why not use `shell.exec('sfdx force:config:get defaultusername');`? and `alias:get` below

test/nut/commands/force/org/status.nut.ts
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as fs from 'fs';
```
WillieRuemmele, line 8: I think this file should become `.sandboxNut.ts` so that it's only ran when we're testing SBs when `[sb-nuts]` is present in a commit, or PR

test/nut/commands/force/org/status.nut.ts
WillieRuemmele, line 107: why not use `shell.exec('sfdx alias:get defaultusername')` like you're doing in the `unsetAlias` method(s) above?

maggiben: Now using sfdx commands

WillieRuemmele (approved):

PR #318, opened 2022-04-27
fix: docs for clean flag

mshanemc:
### What does this PR do?
clarify that org:list --clean only deletes expired scratch orgs

https://github.com/forcedotcom/cli/issues/1449

@W-10886099@

### DIFF
--- a/messages/list.json
+++ b/messages/list.json
@@ -7,7 +7,7 @@
   ],
   "verbose": "list more information about each org",
   "all": "include expired, deleted, and unknown-status scratch orgs",
-  "clean": "remove all local org authorizations for non-active orgs",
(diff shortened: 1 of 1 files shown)


cromwellryan:
This definitely narrows and clarifies the scope of what `--clean` affects. 

The other side of the coin 🪙 is helping the user understand how to deal with the other entries their trying to "clean" up. @jclark-dot-org describes that here:

> I would suggest that if you clarify the docs for --clean as only affecting scratch orgs, you *mention in the same place that auth:logout is the right way to remove non-scratch orgs.*

What would happen if we appended something like:

> remove all local org authorizations for non-active scratch orgs. *`auth:logout` can be used to remove non-scratch orgs*

PR #332, opened 2022-05-20
fix: parking orbit plugin-org

cristiand391:
### What does this PR do?
waiting on https://github.com/forcedotcom/sfdx-core/pull/582 to finish.

@W-10743851@


test/shared/utils.test.ts
```diff
@@ -6,20 +6,22 @@
...
   const sandbox = sinon.createSandbox();
   beforeEach(async () => {
-    stubMethod(sandbox, Aliases, 'create').resolves(Aliases.prototype);
-    stubMethod(sandbox, Aliases, 'getDefaultOptions').returns({});
-    stubMethod(sandbox, Aliases.prototype, 'getKeysByValue')
-      .withArgs('username1')
-      .returns(['alias1'])
-      .withArgs('username2')
-      .returns(['alias2', 'alias2b']);
+    const getAllStub = sandbox.stub();
```
maggiben, line 16: I always like when we use vanilla sinon 👏

src/commands/force/org/beta/create.ts
```diff
@@ -262,9 +261,9 @@ export class Create extends SfdxCommand {
   }
 
   private async setAliasAndDefaultUsername(username: string): Promise<void> {
-    const aliases = await Aliases.create(Aliases.getDefaultOptions());
+    const stateAggregator = await StateAggregator.getInstance();
     if (this.flags.setalias) {
-      await aliases.updateValue(this.flags.setalias, username);
+      stateAggregator.aliases.set(this.flags.setalias, username);
```
maggiben, line 266: I wrote this code and I was forced to use the `updateValue` function, Im glad `set` works just as well

src/commands/force/org/list.ts
```diff
@@ -159,31 +168,40 @@ export class OrgListCommand extends SfdxCommand {
     }
   }
 
-  private getScratchOrgColumnData(): unknown[] {
+  private getScratchOrgColumnData(): Partial<CliUx.Table.table.Columns<Record<string, string>>> {
```
maggiben, line 171: I remember setting the type for column data to `unknown` because it was giving me conflicts with the new table types, I'm glad you solved that, but I get the sense that the type is overly complex for just a simple column data descriptor, I'm sure It won't hurt us in the short~mid term but breaking changes in the library could happen and this is where it gets messy.

maggiben (approved):

maggiben:
- [x] UT pass
- [x] NUTs pass
- [x] Sandbox NUTs pass

src/commands/force/org/list.ts
cristiand391: yep, the whole import looks complex but that's actually just cli-ux after being merge into oclif/core, here's the previous type: https://github.com/salesforcecli/plugin-org/blob/15ae104f9d6be2d1ec1678424db67548312a2ac5/src/commands/force/org/list.ts#L157

PR #358, opened 2022-07-14
fix: set scratch org duration to 7days by default

cristiand391:
### What does this PR do?

Sets `durationdays` flag value to 7 days by default for `force:org:beta:create`, this matches the current org create command from toolbelt.

## Before
the flag value was being passed as undefined to the scratch org payload, then the API would set the duration to 1 day.

## After
if `--durationdays` isn't passed org:beta:create will default to 7 days.

@W-0@

### DIFF
--- a/test/commands/force/org/sandboxCreate.test.ts
+++ b/test/commands/force/org/sandboxCreate.test.ts
@@ -89,7 +89,7 @@ describe('org:create', () => {
     it('will warn the user when --clientid is passed', async () => {
       const commmand = await createCommand(['--type', 'sandbox', '--clientid', '123', '-u', 'testProdOrg']);
       await commmand.runIt();
-      expect(uxWarnStub.calledOnce).to.be.true;
+      expect(uxWarnStub.calledTwice).to.be.true;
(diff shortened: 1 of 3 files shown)


mshanemc (approved):

maggiben (approved):

test/commands/force/org/sandboxCreate.test.ts
```diff
@@ -99,7 +99,7 @@ describe('org:create', () => {
     it('will warn the user when --nonamespace is passed', async () => {
       const commmand = await createCommand(['--type', 'sandbox', '--nonamespace', '-u', 'testProdOrg']);
       await commmand.runIt();
-      expect(uxWarnStub.calledOnce).to.be.true;
+      expect(uxWarnStub.calledTwice).to.be.true;
```
mshanemc, line 102: what caused this to change?

mshanemc (approved):
approved but with a question

test/commands/force/org/sandboxCreate.test.ts
mshanemc: ahh, got it

PR #359, opened 2022-07-28
fix: org list now lists all scratch orgs properly

peternhale:
### What does this PR do?
Fixes the path when reading user json file when the org is referenced in a `00D` file.

@W-11507149@
https://github.com/forcedotcom/cli/issues/1641

### DIFF
--- a/src/shared/orgListUtil.ts
+++ b/src/shared/orgListUtil.ts
@@ -132,9 +132,9 @@ export class OrgListUtil {
           }
           // Theory: within <orgId>.json, if the userId is the first entry, that's the primary username.
           if (orgFileNames.includes(orgFileName)) {
-            const orgFileContent = JSON.parse(
-              await fs.readFile(join(Global.SFDX_STATE_FOLDER, orgFileName), 'utf8')
-            ) as { usernames: string[] };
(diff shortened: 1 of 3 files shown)


src/shared/orgListUtil.ts
```diff
@@ -132,9 +132,9 @@ export class OrgListUtil {
           }
           // Theory: within <orgId>.json, if the userId is the first entry, that's the primary username.
           if (orgFileNames.includes(orgFileName)) {
-            const orgFileContent = JSON.parse(
-              await fs.readFile(join(Global.SFDX_STATE_FOLDER, orgFileName), 'utf8')
-            ) as { usernames: string[] };
+            const orgFileContent = JSON.parse(await fs.readFile(join(Global.SFDX_DIR, orgFileName), 'utf8')) as {
```
peternhale, line 135: Global.SFDX_STATE_FOLDER is a const, where Global.SFDX_DIR is a getter that resolved the state folder path. The cost was causing all reads to fail with file not found.

mshanemc (approved):

mshanemc:
QA notes:
✅ repro of bug (requires user:create) on existing plugin
✅ missing username appears when using this branch

PR #485, opened 2022-11-04
1722: Re-instate lifecycleEventNames so postorgcreate is emitted.

gitfish:
### What does this PR do?
Reinstates lifecycleEventNames prop so that force:org:create will run postorgcreate hooks.

What issues does this PR fix or reference?
[@forcedotcom/cli/issues/1722@](https://github.com/forcedotcom/cli/issues/1722)

### DIFF
--- a/src/commands/force/org/create.ts
+++ b/src/commands/force/org/create.ts
@@ -100,6 +100,7 @@ export class Create extends SfdxCommand {
       description: messages.getMessage('flags.retry'),
     }),
   };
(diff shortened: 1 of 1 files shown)


cristiand391:
**QA notes**:
wrote an oclif plugin to test the hook was being fired
1) Checked out this PR locally
2) built and linked `plugin-org`


✅  Hooks is being fired, can access `postorgcreate` data in the hook:

Screen Shot 2022-11-07 at 11 06 14

PR #517, opened 2022-12-09
fix: always throw errors caught from creating sandboxes

RodEsp:
### What does this PR do?

It ensures that if an error is thrown in an underlying library when attempting to create a Sandbox it will be thrown by this plugin instead of swallowed.

#### Testing Instructions

Before this change:
1. Clone this repo
1. Ensure you're on the `main` branch
1. Run `bin/dev force:org:create -t sandbox sandboxName=testSandbox licenseType=Developer -u <devhub>` against an org that can't create sandboxes such as a basic developer edition org.
1. The command exits with a `0` status code and no error.

After this change:
1. Checkout this branch, `re/return-errors`
1. Run `bin/dev force:org:create -t sandbox sandboxName=testSandbox licenseType=Developer -u <devhub>` against the same org
1. You should see an error that says the following:
```bash
ERROR running force:org:create:  The requested resource does not exist
```

@W-12179113@
@W-12196651@

### DIFF
--- a/src/commands/force/org/create.ts
+++ b/src/commands/force/org/create.ts
@@ -237,8 +237,9 @@ export class Create extends SfdxCommand {
         }
         err.actions = [messages.getMessage('dnsTimeout'), messages.getMessage('partialSuccess')];
         err.exitCode = 68;
-        throw err;
       }
+
+      throw err;
     }
   }
 


WillieRuemmele (approved):
Approved, but can we improve that error message, something like `sandboxes can't be created from this org` right? I remember adding something like that for scratch orgs when the targetdevhub isn't actually a DevHub

RodEsp:
@WillieRuemmele yes, we could improve that error message but I just used that error as an example, this was swallowing _all_ errors that didn't have `'The org cannot be found'` in the message.

WillieRuemmele:
### QA Notes
---
ran the QA notes above and verified
✅ : exits with non-zero exit code
✅ : prints error message (to be improved later)
✅ : will work against a DevHub which can create sandboxes

PR #587, opened 2023-02-20
[DO NOT MERGE] feat: add force:source:open force:mdapi:listmetadata force:mdapi:describemetadata

WillieRuemmele:
### What does this PR do?
Moves `force:source:open` command from plugin-source to `org:open` here
adds new flag, alias, UTs

@W-12566080@
requires: https://github.com/salesforcecli/plugin-source/pull/740

### Release Notes
We changed the names/flags of two commands to be clearer (the old names/flags work fine).
 
`force:mdapi:listmetadata` => `org:list:metadata` 
`--metadatatype` => `--metadata-type`

`force:mdapi:describemetadata` => `org:list:metadata-types`
`-a` for `--api-version` is deprecated.  

For both commands,
`--resultfile` => `--output-file` (still `-f`, no change there!)
`--targetusername`/`-u` => `--target-org`/`-o`

The `force:source:open` command's functionality is being merged with `org:open` as a new flag `--source-file`.  

Because `org:open` is already moved to the new styles, you get the usual `--target-org` and `--api-version` behavior, and `urlonly` becomes `--url-only`.

But since `org:open` has its `--browser` flag, you can now specify which browser your App Builder page opens in.
You can still use `force:source:open` with its original names and flags.


src/commands/org/open.ts
```diff
@@ -41,6 +43,7 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
       char: 'p',
       summary: messages.getMessage('flags.path.summary'),
       env: 'FORCE_OPEN_URL',
+      exclusive: ['source-file'],
```
mshanemc, line 46: nice

src/commands/org/open.ts
```diff
@@ -113,14 +133,35 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
...
+    const accessToken = this.conn.accessToken;
     const instanceUrl = this.org.getField<string>(Org.Fields.INSTANCE_URL);
     const instanceUrlClean = instanceUrl.replace(/\/$/, '');
     return `${instanceUrlClean}/secur/frontdoor.jsp?sid=${accessToken}`;
   }
+
+  private async generateFileUrl(file: string): Promise<string> {
+    try {
+      const metadataResolver = new MetadataResolver();
+      const components = metadataResolver.getComponentsFromPath(file);
```
mshanemc, line 147: I think you might want to assert that exactly one component was found and throw otherwise.

components[0] is like an `undefined` bug waiting to happen

src/commands/org/resume/scratch.ts
```diff
@@ -30,6 +30,7 @@ export default class EnvResumeScratch extends SfCommand<ScratchCreateResponse> {
   public static readonly flags = {
     'job-id': Flags.salesforceId({
       char: 'i',
+      length: 'both',
```
mshanemc, line 33: I'm ok with it as long as it's been tested (automated or manual) to work with both.

mshanemc (approved):
approved with non-blocking comments.

mshanemc (changes requested):
can the open nut use the new flag and do the URL for one or both source-file types

src/commands/org/open.ts
mshanemc: I love `array.at` but we can't use it until node14 is EOL, >4/30.  You'll probably see UT failing from it.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

mshanemc (changes requested):
good except array.at (soon!)

src/commands/org/list/metadata-types.ts
```diff
@@ -0,0 +1,77 @@
...
+const messages = Messages.loadMessages('@salesforce/plugin-org', 'metadata-types');
+
+export class DescribeMetadata extends SfCommand<DescribeMetadataResult> {
+  public static readonly summary = messages.getMessage('description');
+  public static readonly description = messages.getMessage('description');
+  public static readonly examples = messages.getMessages('examples');
+  public static readonly aliases = ['force:mdapi:describemetadata'];
+  public static readonly deprecateAliases = true;
+  public static readonly flags = {
+    'api-version': orgApiVersionFlagWithDeprecations,
```
cristiand391, line 29: The deprecated flag from `sf-plugins-core` doesn't have an alias for `a`:

plugin-source: https://github.com/salesforcecli/plugin-source/blob/main/src/commands/force/mdapi/describemetadata.ts#L27

sf-plugins-core: https://github.com/salesforcecli/sf-plugins-core/blob/main/src/compatibility.ts#L30

Adding it there might cause some conflicts since some commands already use `-a` for other flags:
https://github.com/search?q=org%3Asalesforcecli++%22char%3A+%27a%27%22&type=code

src/commands/org/list/metadata-types.ts
```diff
@@ -0,0 +1,77 @@
...
+    loglevel,
+    'target-org': requiredOrgFlagWithDeprecations,
+    'output-file': Flags.file({
+      aliases: ['resultfile'],
```
cristiand391, line 33: add flag short name `f` to alias:
https://github.com/salesforcecli/plugin-source/blob/main/src/commands/force/mdapi/describemetadata.ts#L32

src/commands/org/list/metadata-types.ts
```diff
@@ -0,0 +1,77 @@
...
+      deprecateAliases: true,
+      char: 'f',
+      description: messages.getMessage('flags.resultfile'),
```
cristiand391, line 36: can we make the message key match the current flag name instead of the deprecated versions?

src/commands/org/list/metadata-types.ts
```diff
@@ -0,0 +1,77 @@
...
+  loglevel,
+  orgApiVersionFlagWithDeprecations,
+  requiredOrgFlagWithDeprecations,
+  SfCommand,
+} from '@salesforce/sf-plugins-core';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/plugin-org', 'metadata-types');
+
+export class DescribeMetadata extends SfCommand<DescribeMetadataResult> {
```
cristiand391, line 22: ditto, class name can match the current command name.

src/commands/org/list/metadata.ts
```diff
@@ -0,0 +1,74 @@
...
+    }),
+    folder: Flags.string({
+      summary: messages.getMessage('flags.folder'),
+      description: messages.getMessage('flagsLong.folder'),
+    }),
+  };
+
+  public async run(): Promise<ListMetadataCommandResult> {
+    const { flags } = await this.parse(ListMetadata);
+    const conn = flags['target-org'].getConnection(flags['api-version']);
```
cristiand391, line 56: in plugin-source this flag was only passed to the `conn.metadata.list` method:
https://github.com/salesforcecli/plugin-source/blob/main/src/commands/force/mdapi/listmetadata.ts#L71

and I think the conn version is being picked from your config values/whatever the default was in sfdxCommand.

Prob don't want to pass it to `org.getConnection` as it may cause a mismatch in apiVersions if users had local/global values.

WillieRuemmele: With `SfCommand`'s calling `.getConnection` is now recommended to pass the api-version flag, even if undefined. I believe that regardless of what version the Connection is created with, it will be able to call `list` with the correct version

cristiand391 (approved):

cristiand391:
**QA notes**:

## `org open`

✅ `force source open` alias
✅ old flag names still work and show deprecation warning:
long:
Screenshot 2023-03-06 at 11 07 19
short:
Screenshot 2023-03-06 at 11 10 23
(`-f` and `-r` doesn't show up as deprecated because the command still use them).

✅ can open a Lightning page in Ligtning App Builder:
```
➜  dreamhouse-lwc git:(main) sf org open --source-file force-app/main/default/flexipages/Broker_Record_Page.flexipage-meta.xml

Opening org **************** as user ********************
Waiting to resolve the Lightning Experience-enabled custom domain...... done
```

✅ opens Lightning App Builder section in Setup if I pass an invalid path to `--source-file` or if SDR doesn't find the type name:
running `sf org open --source-file force-app/main/default/flexipages/lala` opens `https://saas-app-8376-dev-ed.scratch.my.salesforce.com/_ui/flexipage/ui/FlexiPageFilterListPage`.
Previous `force:source:open` implementation throws an error if the path isn't valid.
💡 the command could print a warning (and still open app builder) if the SDR didn't find the type name (or user passed an invalid MD path).

**Note**:
This opens the app builder section in "Salesforce Classic" instead of "Lightning Experience", the command can use this URL to open in Lightning Experience:
https://saas-app-8376-dev-ed.scratch.lightning.force.com/lightning/setup/FlexiPageList/home
instead of:
https://saas-app-8376-dev-ed.scratch.my.salesforce.com/_ui/flexipage/ui/FlexiPageFilterListPage

✅ JSON output matches prev `force:source:open` implementation (`url`, `username` and `orgId` fields in `result` object)

✅ command fails if `--path` and `--source-file` are used at the same time:
```
Error (1): The following errors occurred:
  --path=lightning cannot also be provided when using --source-file
  --source-file=force-app/main/default/flexipages/Broker_Record_Page.flexipage-meta.xml cannot also be provided when using --p
ath
```

⚠️ there's no mention of `--source-file` in the help text/examples.

## `org list metadata`

✅ `force mdapi listmetadata` alias
✅ old flag names still work and show deprecation warning:
long:
```
➜  dreamhouse-lwc git:(main) sf force mdapi listmetadata --apiversion 56.0 --metadatatype FlexiPage --targetusername dreamhouse --logleve
l debug --resultfile output.txt
Warning: The "force mdapi listmetadata" command has been deprecated. Use "org list metadata" instead.
Warning: The "--apiversion" flag has been deprecated. Use "--api-version" instead.
Warning: The "--metadatatype" flag has been deprecated. Use "--metadata-type" instead.
Warning: The "--targetusername" flag has been deprecated. Use "--target-org" instead.
Warning: The loglevel flag is no longer in use on this command. You may use it without error, but it will be ignored.
Set the log level using the `SFDX_LOG_LEVEL` environment variable.
Warning: The "--resultfile" flag has been deprecated. Use "--output-file" instead.
Wrote result file to output.txt.
```
short:
```
➜  dreamhouse-lwc git:(main) ✗ sf force mdapi listmetadata -a 56.0 -m FlexiPage -u dreamhouse --loglevel debug -f output.txt
Warning: The "force mdapi listmetadata" command has been deprecated. Use "org list metadata" instead.
Warning: The "-u" flag has been deprecated. Use "--target-org" instead.
Warning: The loglevel flag is no longer in use on this command. You may use it without error, but it will be ignored.
Set the log level using the `SFDX_LOG_LEVEL` environment variable.
Wrote result file to output.txt.
```
⚠️ there are very few commands that support `-a` for `--api-version` now, since this is just to maintain compatibility we could deprecate `-a` too.

mshanemc (approved):

PR #591, opened 2023-02-22
fix: license-type should not be needed when cloning a sandbox

peternhale:
### What does this PR do?
Fixes issue [#1957](https://github.com/forcedotcom/cli/issues/1957)
@W-12587837@


src/shared/sandboxRequest.ts
```diff
@@ -21,7 +21,7 @@ export const generateSboxName = async (): Promise<string> => {
 };
 
 export async function createSandboxRequest(
-  isClone: true,
+  isClone: true | boolean,
```
mshanemc, line 24: This could be false | undefined then it should work.

src/shared/sandboxRequest.ts
```diff
@@ -21,7 +21,7 @@ export const generateSboxName = async (): Promise<string> => {
 };
 
 export async function createSandboxRequest(
```
mshanemc, line 23: I don't think this overload type change is valid.  If isClone isn't true, would you get back a srcSandboxName like it says ?

src/commands/org/create/sandbox.ts
```diff
@@ -136,10 +136,17 @@ export default class CreateSandbox extends SandboxCommandBase<SandboxProcessObje
 
   private async createSandboxRequest(): Promise<SandboxRequest> {
     // reuse the existing sandbox request generator, with this command's flags as the varargs
-    const { sandboxReq } = await createSandboxRequest(false, this.flags['definition-file'], undefined, {
+    const requestOptions = {
       ...(this.flags.name ? { SandboxName: this.flags.name } : {}),
+      ...(this.flags.clone ? { SourceSandboxName: this.flags.clone } : {}),
       ...(!this.flags.clone && this.flags['license-type'] ? { LicenseType: this.flags['license-type'] } : {}),
```
mshanemc, line 142: the error the user is reporting look like oclif errors because of the oclif `exclusive` aren't they?

mshanemc (changes requested):

src/commands/org/create/sandbox.ts
peternhale: Yes they are exclusive, but I tracked down the error to createSandboxRequest, in the [else block here](https://github.com/salesforcecli/plugin-org/blob/e5b941ade71ab2888b75e3fdff0ffdbc856bb0a1/src/shared/sandboxRequest.ts#L64). `isClone` is false causing the license-type required message in the else block to emit.

Not sure if that code needs to be reworked, but I can see the necessity of the check since this code is once removed from the command.

src/shared/sandboxRequest.ts
peternhale: In this case I need the `true` function signature, given the command is asking for a clone of a sandbox. `org create sandbox` does not use the src sandbox name from the return, but clone must be true in order to run the correct code in the function. This code is also used in the force version of the commands, where that code does need the src sandbox name.

I rechecked the types for each call and type script is telling me that each is getting the expected return type. I guess testing will be the let us know.

src/commands/force/org/clone.ts
```diff
@@ -121,10 +121,10 @@ export class OrgCloneCommand extends SfCommand<SandboxProcessObject> {
       });
 
       const { sandboxReq, srcSandboxName } = await createSandboxRequest(true, flags.definitionfile, logger, varargs);
```
mshanemc, line 123: If you get the types correct, `createSandboxRequest` the compiler will know.

"if true was passed for `isClone` (the first arg) then srcSandboxName will be a string"

That you need a non-null assertion below should tell you that the types are wrong.

src/commands/org/create/sandbox.ts
mshanemc: Won't you need to fix the `exclusive` property on the flags?  Are those errors in the github issue coming from the oclif flag parsing?

peternhale: @mshanemc No, the flags are exclusive and not being used with each other. The error is coming from https://github.com/salesforcecli/plugin-org/blob/main/src/shared/sandboxRequest.ts#L74

src/commands/org/create/sandbox.ts
```diff
@@ -136,10 +136,14 @@ export default class CreateSandbox extends SandboxCommandBase<SandboxProcessObje
...
-    });
+    };
+    const { sandboxReq } = !this.flags.clone
```
WillieRuemmele, line 144: NIT: could simplify this with `await createSandboxRequest(!this.flags.clone, this.flags[...)`

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
✅ : exclusive flags still exclusive
✅ : able to clone SB once removed `--license-type`

before:
```
 ➜  sf org create sandbox --name devsan2  --alias devsan2 --clone WillieUICr -o na40
Error (1): The sandbox license type is required, but you didn't provide a value. Specify the license type in the sandbox definition file with the "licenseType" option, or specify the option as a name-value pair at the command-line. See https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_sandbox_definition.htm for more information.
    SfError: The sandbox license type is required, but you didn't provide a value. Specify the license type in the sandbox definition file with 
    the "licenseType" option, or specify the option as a name-value pair at the command-line. See 
    https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_sandbox_definition.htm for more information.
    Code: 1
```
after
```
 ➜  sf org create sandbox --name devsan2  --alias devsan2 --clone WillieUICr -o na40
=== Config Sandbox Request

 Field       Value              
 ─────────── ────────────────── 
 SandboxName devsan2            
 SourceId    0GQ4p000000HOSwGAO 
 CloneSource WillieUICr         
? Is the configuration correct? Yes
Sandbox Create... ⣟ 00:30:00 until timeout. 0%
```

PR #596, opened 2023-02-26
feat: add namespace column

atsutton:
### What does this PR do?
Adds a 'namespace' column or property to the output of 'org:list' commands. 

Issue #1790: https://github.com/forcedotcom/cli/issues/1790

### DIFF
--- a/src/commands/org/list.ts
+++ b/src/commands/org/list.ts
@@ -121,7 +121,9 @@ export class OrgListCommand extends SfCommand<OrgListResult> {
         .map((org) =>
           Object.fromEntries(
             Object.entries(org).filter(([key]) =>
-              ['defaultMarker', 'alias', 'username', 'orgId', 'connectedStatus'].includes(key)
+              ['defaultMarker', Org.Fields.NAMESPACE_PREFIX, 'alias', 'username', 'orgId', 'connectedStatus'].includes(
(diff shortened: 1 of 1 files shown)


cristiand391:
@atsutton sorry, forgot to assign this in our backlog.
I added this to our current sprint and will review it next week 👍🏼

cristiand391 (commented):
@atsutton I've been testing this and it looks good!

After looking at the code we found that the namespace prop isn't saved to the auth file when creating a scratch org as we thought. The reason it works is because `org list` queries that field here:
https://github.com/salesforcecli/plugin-org/blob/main/src/shared/orgListUtil.ts#L224-L240

The next step for this to ship is to save the org namespace at auth time as mentioned here:
https://github.com/forcedotcom/cli/issues/1790#issuecomment-1303448129
and then update `scratch org create` to properly save the `namespace` prop in the auth file.
(We'll do this work)

We still have to decide if we want to save the namespace as `namespace` for scratch orgs and `namespacePrefix` for dev-ed orgs in the auth files (`org list --json` already returns the namespace for scratch orgs in the `namespace` field so we can't break that).

I'll leave this open until we decide the field name and make the change to get namespaces saved (hopefully next week).
Thanks for your patience!

PR #613, opened 2023-03-15
fix: check if sandbox is resumable

peternhale:
Fixes an improperly formatted message
Bump sfdx-core when [dependent PR is merged](https://github.com/forcedotcom/sfdx-core/pull/787)
@W-12467406@


src/shared/sandboxCommandBase.ts
```diff
@@ -160,8 +160,8 @@ export abstract class SandboxCommandBase<T> extends SfCommand<T> {
       [
         messages.getMessage('sandboxSuccess'),
         messages.getMessages('sandboxSuccess.actions', [
```
cristiand391, line 162: this msg is using a wrong flag:
https://github.com/salesforcecli/plugin-org/blob/main/messages/sandboxbase.md?plain=1#L8

`-e` doesn't exists in `org open`, that should be `--target-org`.

peternhale: @cristiand391 Good catch. Thank you

cristiand391 (approved):

PR #640, opened 2023-04-10
fix: allow expired orgs to be deleted

WillieRuemmele:
### What does this PR do?

allows expired orgs to be deleted 

@W-12992139@
https://github.com/forcedotcom/cli/issues/2045


mshanemc (commented):
does anything similar need to happen for org:delete:sandbox?

mshanemc (approved):

mshanemc:
QA Notes

deleted an expired (deleted) org with force:org:delete -u, and one with -o ✅✅ 
deleted an expired (deleted) org with org:delete:scratch -u, and one with -o ✅✅

put an alias on an expired org, and delete it via force:org:delete ✅  alias is removed from alias:list ✅ 
fine error message when specifying an org with invalid username  ✅  alias is removed from alias:list ✅ 

not tested: deleting sandboxes via force:org:delete

PR #681, opened 2023-05-15
Sm/email-and-source-org

mshanemc:
### What does this PR do?
flags that override the adminEmail and sourceOrg definitionFile props

@W-13191348@
https://github.com/forcedotcom/cli/issues/2130

### DIFF
--- a/src/commands/org/create/scratch.ts
+++ b/src/commands/org/create/scratch.ts
@@ -15,6 +15,7 @@ import {
   SfError,
 } from '@salesforce/core';
 import { SfCommand, Flags } from '@salesforce/sf-plugins-core';
+import { Duration } from '@salesforce/kit';
 import { buildScratchOrgRequest } from '../../../shared/scratchOrgRequest';
 import { buildStatus } from '../../../shared/scratchOrgOutput';
 import { ScratchCreateResponse } from '../../../shared/orgTypes';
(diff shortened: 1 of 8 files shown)


src/commands/org/create/scratch.ts
```diff
@@ -86,6 +84,7 @@ export default class EnvCreateScratch extends SfCommand<ScratchCreateResponse> {
     'duration-days': Flags.duration({
       unit: 'days',
       defaultValue: 7,
+      required: true,
```
shetzel, line 87: Why is this required?

WillieRuemmele (approved):
yay flags.

src/commands/org/create/scratch.ts
mshanemc: good point.  Setting the default the "fixed" way seems to do it better (make it clear that it can't be undefined).  

that didn't used to work correctly, but got fixed (while Mike was in last, I think?)

WillieRuemmele:
### QA Notes
---
✅ : set `adminEmail` in config, create org, click on profile, boom, email matches
✅ : use `--admin-email` flag -> 💥 
✅ : set in config, pass in as flag. flag will override config

✅ : `--source-org` flag validation, length, starts with `00D`
✅ : error when edition is specified and `--source-org`

with dev hub signed up for org snapshot feature:
`sf org create shape -v <devhub>`
`sf org list snapshot` verify it's in output
✅ : `sf org create scratch --source-org 00Dxxx... -v <devhub>`
✅ : `sourceOrg` specified in scratch-def
✅ : flag overrides
✅ : error messages when source org is invalid

PR #716, opened 2023-06-15
Ensure sfdx dir exists

iowillhoit:
### What does this PR do?
Prevents https://github.com/forcedotcom/cli/issues/2222 from happening again.
This was unintentionally fixed in https://github.com/forcedotcom/sfdx-core/pull/842

[@W-13602928@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-13602928)

QA
I reverted back to an older version of the CLI (1.194.1) and then I modified `plugin-org` directly with the same `mkdir` code
Screenshot 2023-06-15 at 3 11 21 PM

Confirm existing directory does not cause issues
Screenshot 2023-06-15 at 3 22 06 PM

### DIFF
--- a/src/shared/orgListUtil.ts
+++ b/src/shared/orgListUtil.ts
@@ -129,6 +129,10 @@ export class OrgListUtil {
    * @param fileNames All the filenames in the global hidden folder
    */
   public static async readAuthFiles(fileNames: string[]): Promise<AuthInfo[]> {
+    // Ensure that the Global.SFDX_DIR exists
+    // https://github.com/forcedotcom/cli/issues/2222
+    await fs.mkdir(Global.SFDX_DIR, { recursive: true });
+
     const orgFileNames = (await fs.readdir(Global.SFDX_DIR)).filter((filename: string) =>
       filename.match(/^00D.{15}\.json$/g)
     );


src/shared/orgListUtil.ts
```diff
@@ -129,6 +129,10 @@ export class OrgListUtil {
    * @param fileNames All the filenames in the global hidden folder
    */
   public static async readAuthFiles(fileNames: string[]): Promise<AuthInfo[]> {
+    // Ensure that the Global.SFDX_DIR exists
+    // https://github.com/forcedotcom/cli/issues/2222
+    await fs.mkdir(Global.SFDX_DIR, { recursive: true });
```
WillieRuemmele, line 134: this'll throw if the directory already exists though, do we need to wrap this in try/catch?

WillieRuemmele (changes requested):

src/shared/orgListUtil.ts
iowillhoit: Oh! I didn't look closely at what you were testing. The `mkdir` function will throw if you are not passing `{ recursive: true }` ([docs](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options:~:text=Calling%20fsPromises.mkdir()%20when%20path%20is%20a%20directory%20that%20exists%20results%20in%20a%20rejection%20only%20when%20recursive%20is%20false)). I just tested this code several more times and it is working as expected.

PR #765, opened 2023-08-04
Sm/sandboxes-in-list

mshanemc:
### What does this PR do?

well, this started off as "make sandboxes a section within org:list" (we can do that more reliably now thanks to Steve's PR for identifying them)

also, while I was in here, I realized **how ugly that table is**.  So

1. remove all the extra ASCII characters.
2. default markers (U) and (D) don't mean anything and don't even align with flags anymore.  Those are now emojis and there's a legend at the bottom of the table.  If you've got better emojideas, holler.
3. One table so the columns align
4. Type field (DevHub, Sandbox, Scratch Org, ?) [the original WI]
5. colors by type, orgs are grouped by those types
6. aliases are at the top of their section instead of the bottom
7. instanceUrl appears when `--verbose` for non-scratch orgs, too
8. long datestamp for createdDate is now just yyyy-mm-dd to match ExpirationDate
9. Column headers are title case instead of UPPERCASE

No JSON should be harmed in the making of this PR

@W-10983482@

### DIFF
--- a/src/commands/org/list.ts
+++ b/src/commands/org/list.ts
@@ -16,9 +16,16 @@ Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/plugin-org', 'list');
 
 export type OrgListResult = {
+  /**
+   * @deprecated
+   * preserved for backward json compatibility.  Duplicates devHubs, sandboxes, regularOrgs, which should be preferred*/
   nonScratchOrgs: ExtendedAuthFields[];
   scratchOrgs: FullyPopulatedScratchOrgFields[];
+  sandboxes: ExtendedAuthFields[];
+  regularOrgs: ExtendedAuthFields[];
+  devHubs: ExtendedAuthFields[];
 };
+
 export class OrgListCommand extends SfCommand<OrgListResult> {
   public static readonly summary = messages.getMessage('summary');
   public static readonly examples = messages.getMessages('examples');
@@ -56,6 +63,9 @@ export class OrgListCommand extends SfCommand<OrgListResult> {
     this.flags = flags;
     const metaConfigs = await OrgListUtil.readLocallyValidatedMetaConfigsGroupedByOrgType(fileNames, flags);
     const groupedSortedOrgs = {
+      devHubs: metaConfigs.devHubs.map(decorateWithDefaultStatus).sort(comparator),
+      regularOrgs: metaConfigs.regularOrgs.map(decorateWithDefaultStatus).sort(comparator),
+      sandboxes: metaConfigs.sandboxes.map(decorateWithDefaultStatus).sort(comparator),
       nonScratchOrgs: metaConfigs.nonScratchOrgs.map(decorateWithDefaultStatus).sort(comparator),
       scratchOrgs: metaConfigs.scratchOrgs.map(decorateWithDefaultStatus).sort(comparator),
       expiredScratchOrgs: metaConfigs.scratchOrgs.filter((org) => !identifyActiveOrgByStatus(org)),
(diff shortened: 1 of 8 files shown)


src/commands/org/list.ts
```diff
@@ -8,17 +8,28 @@
...
+const defaultHubEmoji = '🌳';
+
 export type OrgListResult = {
+  /**
+   * @deprecated
+   * preserved for backward json compatibility.  Duplicates devHubs, sandboxes, regularOrgs, which should be preferred*/
   nonScratchOrgs: ExtendedAuthFields[];
   scratchOrgs: FullyPopulatedScratchOrgFields[];
+  sandboxes: ExtendedAuthFields[];
+  regularOrgs: ExtendedAuthFields[];
```
WillieRuemmele, line 29: what are "regular" orgs?

src/commands/org/list.ts
```diff
@@ -70,15 +84,30 @@ export class OrgListCommand extends SfCommand<OrgListResult> {
...
 
-    this.printOrgTable(result.nonScratchOrgs, flags['skip-connection-status']);
+    this.printOrgTable({
+      devHubs: result.devHubs,
+      regularOrgs: result.regularOrgs,
+      sandboxes: result.sandboxes,
+      scratchOrgs: result.scratchOrgs,
+      skipconnectionstatus: flags['skip-connection-status'],
+    });
+    // this.printScratchOrgTable(result.scratchOrgs);
```
WillieRuemmele, line 103: can remove

src/commands/org/list.ts
```diff
@@ -112,108 +141,96 @@ export class OrgListCommand extends SfCommand<OrgListResult> {
...
+const statusToEmoji = <T extends ExtendedAuthFields | FullyPopulatedScratchOrgFields>(val: T): T => ({
+  ...val,
+  defaultMarker: val.defaultMarker?.replace('(D)', defaultHubEmoji)?.replace('(U)', defaultOrgEmoji),
 });
 
 // sort by alias then username
 const comparator = <T extends ExtendedAuthFields | FullyPopulatedScratchOrgFields>(a: T, b: T): number => {
-  const aliasCompareResult = (a.alias ?? '').localeCompare(b.alias ?? '');
-  return aliasCompareResult !== 0 ? aliasCompareResult : (a.username ?? '').localeCompare(b.username);
+  const emptiesLast = Array(10).fill('z').join('');
```
WillieRuemmele, line 231: NIT: lots of iterations just to fill an array with 10 'z's.. could hard code this and save _some_ time... but that's why it's a NIT 😄

WillieRuemmele (changes requested):

src/commands/org/list.ts
mshanemc: not scratch, sandbox, or devHubs.  got a better name?

src/commands/org/list.ts
mshanemc: yeah

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
✅ : everything demo'd in demos
✅ : don't have a sandbox - saw it colored and delineated in demos
✅ : `--clean` / `--all` / `--no-prompt` behavior
✅ : emojis in terminal/macOS at least 🤷‍♂️ 
🏆 : colors
🏆 : aliases first
🏆 : less cluttered table
🏆 : more consistent formats
✅ : `--json` behavior 
```bash
 ➜  sfdx org:list --all --json | jq -r '.result.nonScratchOrgs | length'
2
➜  dreamhouse-lwc git:(main) ✗  hub:(DevHub) scratch:(test-ewtog37qdlga@example.com)
 ➜  ../../oss/plugin-org/bin/dev org:list --all --json | jq -r '.result.nonScratchOrgs | length' 
2
```

PR #857, opened 2023-11-06
fix: check for both sf and sfdx env vars

shetzel:
### What does this PR do?
Check for both `SF_CONTAINER_MODE` and `SFDX_CONTAINER_MODE`
Check for both `SF_DOMAIN_RETRY` and `SFDX_DOMAIN_RETRY`

@W-14437987@
#2554

### DIFF
--- a/src/commands/org/open.ts
+++ b/src/commands/org/open.ts
@@ -72,6 +72,7 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
     this.conn = this.org.getConnection(flags['api-version']);
 
     let url = await this.buildFrontdoorUrl();
+    const env = new Env();
 
     if (flags['source-file']) {
       url += `&retURL=${await this.generateFileUrl(flags['source-file'])}`;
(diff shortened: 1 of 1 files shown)


src/commands/org/open.ts
```diff
@@ -115,7 +117,8 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
       if (err instanceof Error) {
         if (err.message.includes('timeout')) {
           const domain = `https://${/https?:\/\/([^.]*)/.exec(url)?.[1]}.lightning.force.com`;
-          const timeout = new Duration(new Env().getNumber('SFDX_DOMAIN_RETRY', 240), Duration.Unit.SECONDS);
+          const domainRetryTimeout = env.getNumber('SF_DOMAIN_RETRY') ?? env.getNumber('SFDX_DOMAIN_RETRY', 240);
```
WillieRuemmele, line 120: should this be the same `||`?

shetzel: No because `env.getNumber()` does not have a default value (like getBoolean() does) so if the env var is not set it returns undefined.

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
✅ `SF/DX_CONTAINER_MODE=true` -> printed warning correctly
✅ `SF/DX_DOMAIN_RETRY =true` -> set timeout correctly

PR #868, opened 2023-11-17
Sm/tracking-management

mshanemc:
### What does this PR do?
2 new commands
`sf org enable tracking`
`sf org disable tracking`

@W-14513033@
roundabout fix for https://github.com/forcedotcom/cli/issues/2495

QA:

cat your `~/.sfdx/` as you use these.

Nuts cover the "can't enable tracking on the devHub" case, as well as idempotency.

check messages carefully

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,10 +72,10 @@
             "description": "Delete scratch orgs, sandboxes, org shapes, and org snapshots."
           },
           "enable": {
-            "description": "description for org.enable"
+            "description": "Enable tracking on an org."
           },
           "disable": {
-            "description": "description for org.disable"
+            "description": "Disable tracking on an org"
           }
         },
         "external": true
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -70,8 +70,15 @@
           },
           "delete": {
             "description": "Delete scratch orgs, sandboxes, org shapes, and org snapshots."
+          },
+          "enable": {
+            "description": "description for org.enable"
```
mdonnalley, line 75: these need to be updated

mdonnalley (changes requested):

package.json
mshanemc: good catch.

mdonnalley (approved):

mdonnalley:
QA:

🟢 `sf org disable tracking` on scratch org sets `tracksSource` to false in auth file
🟢 `sf org enable tracking` on scratch org sets `tracksSource` to true in auth file
🟢 `sf org enable tracking` on devhub fails

PR #877, opened 2023-12-01
feat: add namespace column (#596)

cristiand391:
### What does this PR do?
Adds `namespace` column to `sf org list --verbose`.

@W-12009216@


cristiand391:
✅ json still includes namespace if `--verbose` is missing
✅ only render `namespace` column if `--verbose`

PR #894, opened 2023-12-15
Sm/missing-deps

mshanemc:
### What does this PR do?
adds missing dependencies
note: chalk has a higher ESM major version than 4, but we can't use it becuase there's exported bits of chalk (standardColors) from sf-plugins-core and they're not compatible (api changes).

We'll need to use ESM chalk in sf-plugins-core and then could bump here and elsewhere

https://github.com/forcedotcom/eslint-config-salesforce-typescript/actions/runs/7214346950/job/19656237273

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,8 @@
     "@salesforce/kit": "^3.0.15",
     "@salesforce/sf-plugins-core": "^5.0.3",
     "@salesforce/source-deploy-retrieve": "^10.0.3",
+    "chalk": "^4",
+    "change-case": "^5.3.0",
     "open": "^9.1.0"
   },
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -10,6 +10,8 @@
     "@salesforce/kit": "^3.0.15",
     "@salesforce/sf-plugins-core": "^5.0.3",
     "@salesforce/source-deploy-retrieve": "^10.0.3",
+    "chalk": "^4",
+    "change-case": "^5.3.0",
```
cristiand391, line 14: same comment about change-case:
https://github.com/salesforcecli/plugin-data/pull/772#discussion_r1428099925

PR #913, opened 2024-01-02
feat: frontdoor via post

mshanemc:
### What does this PR do?
- keep tokens out of urls for human open
- create a small html file locally that redirects to frontdoor.jsp via post
- open that local file in the browser

@W-10025523@

also

What does this PR do?
add browserPrivate as an option for --browser

What issues does this PR fix or reference?
@W-14764106@
https://github.com/forcedotcom/cli/discussions/2262

https://github.com/forcedotcom/cli/issues/1852

https://github.com/forcedotcom/cli/issues/636

### Notes
it is valid to do `--json`.  That'll still have a AT-param in the url for machine use AND do the new open behavior for humans

### QA challenges

- does it work correctly on windows (that is, fs and tempDir are working correctly) and across various browsers and shells?
- can you find a way to leave these temporary files on the filesystem with valid AT in them?


src/commands/org/open.ts
```diff
@@ -130,48 +123,21 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
...
+    const openOptions =
+      // assertions could be removed once oclif flag typins are fixed
+      flags.browser ? { app: { name: apps[flags.browser as 'chrome' | 'edge' | 'firefox'] } } : {};
 
-    await utils.openUrl(url, openOptions);
-    return output;
-  }
+    // create a local html file that contains the POST stuff.
+    // for review...there's always an access token, right?
+    const tempFilePath = path.join(tmpdir(), `org:open-${new Date().valueOf()}.html`);
```
cristiand391, line 132: Replace colons in filename with hyphen or underscore, colons aren't allowed in file paths in Windows
https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

src/commands/org/open.ts
```diff
@@ -130,48 +123,21 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
       throw err;
     }
 
-    const openOptions = flags.browser
-      ? // assertion can be removed once oclif option flag typings are fixed
-        { app: { name: apps[flags.browser as 'chrome' | 'edge' | 'firefox'] } }
-      : {};
+    const openOptions =
+      // assertions could be removed once oclif flag typins are fixed
+      flags.browser ? { app: { name: apps[flags.browser as 'chrome' | 'edge' | 'firefox'] } } : {};
```
cristiand391, line 128: _I think_ we can remove the assertion if we make use the new `Flags.option` for `flags.browser`.

src/commands/org/open.ts
```diff
@@ -130,48 +123,21 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
...
-      } else {
-        return 'lightning/setup/FlexiPageList/home';
-      }
-    } catch (error) {
-      if (error instanceof Error && error.name === 'FlowIdNotFoundError') {
-        this.error(error);
-      }
-      return 'lightning/setup/FlexiPageList/home';
-    }
+    await rm(tempFilePath, { force: true });
```
cristiand391, line 139: if there's an error while deleting file (slow machine, browser still using it after 1s?) `rm` will not retry the operation:
https://nodejs.org/docs/latest-v20.x/api/fs.html#fspromisesrmpath-options

let's set a `maxRetries` just to be safe. I think 3 should be good, will try to break it on QA.

cristiand391 (changes requested):
Looks good, just left 1 suggestion and 2 requested changes.

src/commands/org/open.ts
mshanemc: I closed 480.  I'll move this flag change idea to https://github.com/salesforcecli/plugin-org/pull/917

src/commands/org/open.ts
mshanemc: good idea.  Looks like the retries needs `recursive:true`

cristiand391 (approved):

cristiand391:
QA notes:

testing done on win10 vm:

❌ tmp file gets deleted too quickly, tried ~20 times with firefox/edge, with them open/closed but by the time they try to access the file it is gone:
Screenshot 2024-01-09 at 14 23 10
Screenshot 2024-01-09 at 14 23 24
Screenshot 2024-01-09 at 14 23 52
Screenshot 2024-01-09 at 14 24 05

### NOTE

edited the code to print the tmp file path.
sleeping for 3s works sometimes, but found 5s to be most consistent.

having the browser closed didn't seem to make any difference (I guess b/c the child procecss gets returned once it's opened so the time starts to run after that happens)

If I. delete the `await rm` line it works on firefox/edge and pwsh/git-bash/cmd

cristiand391:
QA update:

### windows

✅ browsers can open the file and do auth

`sf org open -o cd --browser <browser>`, without browser flag it still opens the default.
tested with firefox/edge and latest chrome.
Added a console.log to get the tmp filepath, always get deleted after the sleep (more details below)

✅ file gets deleted even if you specify a browser that isn't installed
uninstalled chrome and ran `sf org open -o cd --browser chrome`, the tmp file was created and deleted after 7s, exit code 0.

✅ file gets deleted if I kill the browser before the 7s wait


✅ `--json` withouth `--url-only`
sf opens the tmp file for auth but still includes the frontdoor URL + AT in json. 
json doesn't include path to tmp file
file gets deleted

✅ `--json` with `--url-only`
doesn't open browser/doesn't create tmp file. frontdoor URL + AT in json

messages/open.md
```diff
@@ -12,12 +12,18 @@ To generate a URL but not launch it in your browser, specify --url-only.
 
 To open in a specific browser, use the --browser flag. Supported browsers are "chrome", "edge", and "firefox". If you don't specify --browser, the org opens in your default browser.
 
+To open in an incognito window using your default browser use `--browser browserPrivate`
```
jshackell-sfdc, line 15:
```suggestion
To open in an incognito window using your default browser, use `--browser browserPrivate`. 
```

messages/open.md
```diff
@@ -12,12 +12,18 @@ To generate a URL but not launch it in your browser, specify --url-only.
...
+
 # examples
 
 - Open your default org in your default browser:
 
   $ <%= config.bin %> <%= command.id %>
 
+- Open your default org in your default browser using an incognito window
```
jshackell-sfdc, line 23:
```suggestion
- Open your default org in an incognito window of your default browser:
```

jshackell-sfdc (changes requested):

src/commands/org/open.ts
```diff
@@ -36,7 +38,7 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
     browser: Flags.option({
       char: 'b',
       summary: messages.getMessage('flags.browser.summary'),
-      options: ['chrome', 'edge', 'firefox'] as const, // These are ones supported by "open" package
+      options: ['chrome', 'edge', 'firefox', 'browserPrivate'] as const, // These are ones supported by "open" package
```
cristiand391, line 41: I understand this follows the `open` API but not sure if we should expose it in the CLI.

what if you add a new `--private` flag that's exclusive with `--browser`?

src/commands/org/open.ts
```diff
@@ -128,47 +124,20 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
...
-    await utils.openUrl(url, {
+    // create a local html file that contains the POST stuff.
+    // for review...there's always an access token, right?
+    const tempFilePath = path.join(tmpdir(), `org-open-${new Date().valueOf()}.html`);
+    await writeFile(tempFilePath, getFileContents(conn.accessToken as string, conn.instanceUrl, retUrl));
+    await utils.openUrl(`file:///${tempFilePath}`, {
       ...(flags.browser ? { app: { name: apps[flags.browser] } } : {}),
-      // only applies on macOS.  open library ignores this property as it is the default behavior on other OS
-      newInstance: true,
+      ...(flags.browser === 'browserPrivate' ? { newInstance: true } : {}),
```
cristiand391, line 133: `newInstance` is a macos-only feature:
https://github.com/sindresorhus/open?tab=readme-ov-file#newinstance-macos-only

for private you need to use `apps.browserPrivate`:
https://github.com/sindresorhus/open?tab=readme-ov-file#usage

cristiand391 (changes requested):
1 suggestion about flags and 1 requested change for `open`'s incognito API

src/commands/org/open.ts
mshanemc: I almost did it that way.  I'll let your vote tip it to that side.

src/commands/org/open.ts
mshanemc: yes, but
1. it has no effect when used on other platforms (per `open`'s code)
2. we only need it when doing private (otherwise private doesn't work)

Are you wanting the criteria to be tighter, like... if macos AND private, then `newInstance:true` ?

cristiand391: ahh right, nvm then, I see you commented this in your previous incognito PR.

cristiand391 (approved):

PR #934, opened 2024-01-16
fix: decode `path` flag before writing it to form

cristiand391:
### What does this PR do?
Follow up from https://github.com/salesforcecli/plugin-org/pull/913

`sf org open --path <path>` isn't working on current nightly (sf v2.25.5).

Screenshot 2024-01-16 at 12 14 04

`--path` is URI-encoded by default (for the old behavior of opening the url with the AT + path | source-file as param), but the form redirect requires is to be decoded:

Screenshot 2024-01-16 at 12 15 19

tested on macos (firefox/chrome) and windows (firefox/edge)

[skip-validate-pr]

### DIFF
--- a/src/commands/org/open.ts
+++ b/src/commands/org/open.ts
@@ -130,7 +130,16 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
 
     // create a local html file that contains the POST stuff.
     const tempFilePath = path.join(tmpdir(), `org-open-${new Date().valueOf()}.html`);
-    await fs.promises.writeFile(tempFilePath, getFileContents(conn.accessToken as string, conn.instanceUrl, retUrl));
+    await fs.promises.writeFile(
+      tempFilePath,
+      getFileContents(
+        conn.accessToken as string,
+        conn.instanceUrl,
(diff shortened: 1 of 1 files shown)


src/commands/org/open.ts
```diff
@@ -130,7 +130,16 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
...
     const tempFilePath = path.join(tmpdir(), `org-open-${new Date().valueOf()}.html`);
-    await fs.promises.writeFile(tempFilePath, getFileContents(conn.accessToken as string, conn.instanceUrl, retUrl));
+    await fs.promises.writeFile(
+      tempFilePath,
+      getFileContents(
+        conn.accessToken as string,
+        conn.instanceUrl,
+        // the path flag is URI-encoded in its `parse` func.
+        // For the form redirect to work we need it decoded.
+        flags.path ? decodeURIComponent(flags.path) : retUrl
```
cristiand391, line 140: `retUrl` can be either a generated path from `--source-file` or `--path`.
If the former that's fine because it doesn't get encoded in `generateFileUrl`, so we only check for `flags.path`.

mshanemc (approved):

PR #962, opened 2024-02-22
fix: convert linux file path to Windows one in WSL

nrakuyama:
### What does this PR do?

Make `org open` command to convert Linux file path to Windows one in WSL.

`org open` command was changed to open temp html file from to open url directly.
In WSL, the command silently fails because Windows browser tries to open linux file path.

https://github.com/forcedotcom/cli/issues/2677
[@W-15278875@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-15278875)

### DIFF
--- a/src/commands/org/open.ts
+++ b/src/commands/org/open.ts
@@ -8,6 +8,7 @@
 import path from 'node:path';
 import { platform, tmpdir } from 'node:os';
 import fs from 'node:fs';
+import { execSync } from 'node:child_process';
 import {
   Flags,
   loglevel,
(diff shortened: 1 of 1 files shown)


iowillhoit:
Hey @nrakuyama, I do not personally have a way to confirm this is working as expected. From some searching, it looks like `wslpath` is not a built-in utility, but needs to be installed. What happens with this code if the user does not have `wslpath` installed? I assume it fails, but does the error message make it clear that the user needs to install the utility?

nrakuyama:
Hi, @iowillhoit 
`wslpath` command is bulit into WSL and WSL users don't need to install it.
It was added from Build 17046 which is April 2018 update.
https://learn.microsoft.com/en-us/windows/wsl/release-notes

If `wslpath` is not installed, command not found error message will appear.
image

iowillhoit:
Thanks for the confirmation and the screenshot @nrakuyama, I just confirmed that this is still working as expected locally 👍

iowillhoit (approved):

PR #992, opened 2024-03-21
Use homefolder for `org open` under linux to prevent issues with sandboxing through snap

fmannreply:
### What does this PR do?
Currently `sf org open` creates a file under `/tmp` (base on TMPDIR/TMP environment vars). The `/tmp` folder is not visible to snaps due to sandboxing. Snaps are the recommended distribution for Firefox under Ubuntu.

Instead of creating a file under `tmp` this creates the file under the homefolder in a subdirectory called `temp`. `/home/.tmp` also does not work, as snaps can not see hidden folders. The temp files could also be created under `/home` directly, as they are deleted either way.

https://github.com/forcedotcom/cli/issues/2769

### DIFF
--- a/src/commands/org/open.ts
+++ b/src/commands/org/open.ts
@@ -130,7 +130,15 @@ export class OrgOpenCommand extends SfCommand<OrgOpenOutput> {
     }
 
     // create a local html file that contains the POST stuff.
-    const tempFilePath = path.join(tmpdir(), `org-open-${new Date().valueOf()}.html`);
+    let tempFilePath = path.join(tmpdir(), `org-open-${new Date().valueOf()}.html`);
+
+    if( platform() === 'linux'){
+        // on linux create the temp file in the home folder as e.g. snaps can not access /tmp
+        let folderPath = path.join(homedir(), `temp`);
+        await fs.promises.mkdir(folderPath, { recursive: true });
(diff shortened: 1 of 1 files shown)


iowillhoit:
Hey @fmannreply, I don't think we would want to make this change for all Linux machines.

You should be able to set your temp directory for this command. Try running `TMPDIR="$HOME/temp" sf org open [your-flags]`

fmannreply:
Hi @iowillhoit , yep your mentioned workaround works. Adding the environment var under `bashrc` should also provide a permanent fix for me.

However, this behavior effects all linux users using snap, which is the default distribution on Ubuntu. I also checked on a new release (23.10) and there the tmp folder also defaults to `/tmp`.

I can still understand if you don't want to make the changes for all Linux machines.

iowillhoit:
Hey @fmannreply thanks for the reply, I am glad that workaround worked for you. 

I spoke with the team and we are not going to make this change for Linux. The main reason for this decision is that this particular piece of functionality is temporary, and will be removed in the future as we collaborate with another team at Salesforce. Setting `TMPDIR` is what we will suggest as a workaround until it is removed. 

Thank you so much for being willing to open a Pull Request. We really do appreciate it!

PR #1021, opened 2024-04-09
fix: typo in description

jshackell-sfdc:
### What does this PR do?

Fixes typo in description that says "production org" instead of "source org". 

@W-15458307@

### DIFF
--- a/messages/refresh.sandbox.md
+++ b/messages/refresh.sandbox.md
@@ -4,7 +4,7 @@ Refresh a sandbox org using the sandbox name.
 
 # description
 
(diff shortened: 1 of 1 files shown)


jshackell-sfdc:
@shetzel -- super easy PR for your review!  One question, though: is it still correct to refer to the org that has the sandbox licenses as the "production org"?  I think so, but please confirm.  I ask 'cause that term is still used in the help, but in that context. Thx.

shetzel:
> @shetzel -- super easy PR for your review! One question, though: is it still correct to refer to the org that has the sandbox licenses as the "production org"? I think so, but please confirm. I ask 'cause that term is still used in the help, but in that context. Thx.

Yes, that's correct.

shetzel (approved):

PR #1119, opened 2024-06-27
fix: ensure org display picks the correct record

cristiand391:
### What does this PR do?

Fix a bug where `org display` would return wrong info about a scratch org when its 15-char ID matches other `ScratchOrgInfo` records in the hub.

the query runs here:
https://github.com/salesforcecli/plugin-org/blob/49c4e33cf3a612b4f2344f0d0fe11aa0de3e223a/src/shared/orgListUtil.ts#L244C1-L251C43

and gets translated to this SOQL query:
```sql
SELECT CreatedDate, Edition, Status, ExpirationDate, Namespace, OrgName, CreatedBy.Username, SignupUsername FROM ScratchOrgInfo WHERE ScratchOrg IN ('org-id')
```

The `IN` operator is case-sensitive if the fields it operates on are, `ScratchOrgInfo.ScratchOrg` **is not**:
```
sf sobject describe --sobject ScratchOrgInfo --target-org na40 --json | jq '.result.fields | .[] | select (.name == "ScratchOrg") | .caseSensitive'

> false
```

https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_conditionexpression.htm
> Operators that compare values, such as =,<=, IN, and LIKE. Operators are case insensitive for most fields, but case sensitive for case-sensitive fields.

We now look in the array of returned records for the one that matches the username from the auth file.

Slack thread: https://salesforce-internal.slack.com/archives/G02K6C90RBJ/p1719501479180489

## Testing

You'll need a scratch org with an ID that matches other records in the hub, since that's out of control you can partially repro this by creating a scratch org  using the na40 hub and then modify its auth file and set this org ID:
```
00D8K0000016nErUAI
```

running `sf org display` will show a deleted scratch org with this data:
```
...
 Signup Username test-taa4gxvtq9cc@example.com
 Status          Deleted
...
```
note that the signup username is different from the one in your auth file (printed a few lines below as `username`)

with this branch linked, `org display` will fail saying it can't find the org because the record doesn't match with your local username.

There's also a UT that covers this scenario.

@W-16114810@

### DIFF
--- a/src/commands/org/display.ts
+++ b/src/commands/org/display.ts
@@ -56,7 +56,8 @@ export class OrgDisplayCommand extends SfCommand<OrgDisplayReturn> {
     const fields = authInfo.getFields(true) as AuthFieldsFromFS;
 
     const isScratchOrg = Boolean(fields.devHubUsername);
-    const scratchOrgInfo = isScratchOrg && fields.orgId ? await this.getScratchOrgInformation(fields.orgId) : {};
+    const scratchOrgInfo =
+      isScratchOrg && fields.orgId ? await this.getScratchOrgInformation(fields.orgId, fields.username) : {};
 
     const returnValue: OrgDisplayReturn = {
       // renamed properties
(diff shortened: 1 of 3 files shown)


src/commands/org/display.ts
```diff
@@ -101,11 +102,15 @@ export class OrgDisplayCommand extends SfCommand<OrgDisplayReturn> {
...
-  private async getScratchOrgInformation(orgId: string): Promise<ScratchOrgFields> {
+  private async getScratchOrgInformation(orgId: string, username: string): Promise<ScratchOrgFields> {
     const hubOrg = await this.org.getDevHubOrg();
     // we know this is a scratch org so it must have a hubOrg and that'll have a username
     const hubUsername = hubOrg?.getUsername() as string;
-    const result = (await OrgListUtil.retrieveScratchOrgInfoFromDevHub(hubUsername, [trimTo15(orgId)]))[0];
+    // this query can return multiple records that match the 15 char ID because `ScratchOrgInfo.ScratchOrg` isn't a case-sensitive field
+    // so we look for the record that matches the scratch org username in the auth file.
+    const result = (await OrgListUtil.retrieveScratchOrgInfoFromDevHub(hubUsername, [trimTo15(orgId)])).find(
+      (rec) => rec.SignupUsername === username
```
cristiand391, line 112: `OrgListUtil.retrieveScratchOrgInfoFromDevHub` is also used by `org list` but it's not a problem there because table rows data is generated from each payload.

`org display` builds the info object for one org by merging auth fields and the payload response so we need to be sure it's the correct record here.

src/shared/orgTypes.ts
```diff
@@ -52,7 +52,6 @@ export type FullyPopulatedScratchOrgFields = ScratchOrgFields &
 
 // developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_scratchorginfo.htm
 export type ScratchOrgInfoSObject = {
-  Id: string;
   CreatedDate: string;
```
cristiand391, line 55: I don't see this `Id` field either in docs or the API response. The org ID returned by `org display` is comes from the auth file.

mdonnalley (approved):

PR #1141, opened 2024-07-23
feat: list metadata/-types stdout is table

WillieRuemmele:
### What does this PR do?

moves regular output into a table for 
`list metadata` 
`list metadata-types`

@W-6015646@

### DIFF
--- a/src/commands/org/create/sandbox.ts
+++ b/src/commands/org/create/sandbox.ts
@@ -216,14 +216,10 @@ export default class CreateSandbox extends SandboxCommandBase<SandboxCommandResp
 
     const data = Object.entries(sandboxReq).map(([key, value]) => ({ key, value }));
     this.styledHeader('Config Sandbox Request');
-    this.table(
-      data,
-      {
-        key: { header: 'Field' },
-        value: { header: 'Value' },
-      },
-      {}
-    );
+    this.table(data, {
+      key: { header: 'Field' },
+      value: { header: 'Value' },
+    });
 
     if (
       !(await this.confirm({
--- a/src/commands/org/create/sandbox.ts
+++ b/src/commands/org/create/sandbox.ts
@@ -8,7 +8,6 @@
 import { Duration } from '@salesforce/kit';
 import { Flags } from '@salesforce/sf-plugins-core';
 import { Lifecycle, Messages, SandboxEvents, SandboxRequest, SfError } from '@salesforce/core';
-import { Ux } from '@salesforce/sf-plugins-core';
 import { Interfaces } from '@oclif/core';
 import requestFunctions from '../../../shared/sandboxRequest.js';
 import { SandboxCommandBase, SandboxCommandResponse } from '../../../shared/sandboxCommandBase.js';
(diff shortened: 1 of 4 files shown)


src/commands/org/list/metadata-types.ts
```diff
@@ -68,7 +68,18 @@ export class ListMetadataTypes extends SfCommand<DescribeMetadataResult> {
       await fs.promises.writeFile(flags['output-file'], JSON.stringify(describeResult, null, 2));
       this.logSuccess(`Wrote result file to ${flags['output-file']}.`);
     } else {
-      this.styledJSON(describeResult);
+      this.table(
+        describeResult.metadataObjects,
+        Object.fromEntries(Object.keys(describeResult.metadataObjects[0]).map((k) => [k, { header: k }])),
```
mshanemc, line 73: the type says that suffix is potentially undefined.  If that's true (?), and the first item didn't have a suffix, then the suffix column wouldn't exist even if other items did have one.

src/commands/org/create/sandbox.ts
```diff
@@ -215,14 +214,16 @@ export default class CreateSandbox extends SandboxCommandBase<SandboxCommandResp
...
     this.styledHeader('Config Sandbox Request');
-    this.table(data, columns, {});
+    this.table(
+      data,
+      {
+        key: { header: 'Field' },
+        value: { header: 'Value' },
+      },
+      {}
+    );
```
mshanemc, line 226:
```suggestion
    this.table(
      data,
      {
        key: { header: 'Field' },
        value: { header: 'Value' },
      }
    );
```

src/commands/org/refresh/sandbox.ts
```diff
@@ -249,14 +248,16 @@ export default class RefreshSandbox extends SandboxCommandBase<SandboxCommandRes
...
     this.styledHeader('Config Sandbox Refresh');
-    this.table(data, columns, {});
+    this.table(
+      data,
+      {
+        key: { header: 'Field' },
+        value: { header: 'Value' },
+      },
+      {}
+    );
```
mshanemc, line 260:
```suggestion
    this.table(
      data,
      {
        key: { header: 'Field' },
        value: { header: 'Value' },
      }
    );
```

src/commands/org/list/metadata.ts
```diff
@@ -62,7 +62,11 @@ export class ListMetadata extends SfCommand<ListMetadataCommandResult> {
       fs.writeFileSync(flags['output-file'], JSON.stringify(listResult, null, 2));
       this.logSuccess(`Wrote result file to ${flags['output-file']}.`);
     } else if (listResult?.length) {
-      this.styledJSON(listResult);
+      this.table(listResult, Object.fromEntries(Object.keys(listResult[0]).map((k) => [k, { header: k }])), {
```
mshanemc, line 65: FileProperties also has 2 potentially undefined fields.  If the first item doesn't have them, those columns will be missing from all the other items that might

mshanemc (changes requested):
I think you gotta choose between hardcoding the columns OR scanning the full results set to build the Set of fields for the columns.

They could share a function to do that.

and then I think you probably want to capitalize them and split with spaces to make it look nicer than

createdById: string;
    createdByName: string;
    createdDate: string;
    fileName: string;
    fullName: string;
    id: string;
    lastModifiedById: string;
    lastModifiedByName: string;
    lastModifiedDate: string;
    manageableState?: string | null | undefined;
    namespacePrefix?: string | null | undefined;
    type: string;

wouldl provide

mshanemc:
QA:
`org list metadata-types -o hub`
✅ works, columns look fine
🎨  I think these should be sorted by Xml Names.  I'd also maybe alphabetize the childXmlNames (nit)

`./bin/run.js org list metadata -o hub -m CustomObject`
✅ works
❌  `type` column is lowercased.  I'm not sure it should be there at all (is there a case where it wouldn't just be the type you just asked for?)

💡 I only see `Id` and `Namespace Prefix` when there's a `Manageable State`.  I wonder if there should be 2 or 3 tables, grouped by Manageable state, so that the table without manageable state doesn't need those fields?

🎨 I think a reasonable sort order for these would be state, prefix (unless you decide to do separate tables), full name.
🎨 I'm not a fan of the 1970-01-01.  I think I'd rather those be empty than "obviously wrong date."  And they really weren't Created or Modified on that date by somebody, either.  They were just Standard Objects that predate the org?

mshanemc (approved):

PR #1396, opened 2025-04-10
fix: W-18237102 - refresh sandbox with sourceSandboxName

soridalac:
### What does this PR do?
fix issue: running` sf org refresh sandbox --name sbx --target-org prodOrgAlias` using `sourceSandboxName` to clone from different existing sandbox production.

it works with ` sourceId` and not `sourceSandboxName`. 

Add flag feature: `'source-id'` and `'source-sandbox-name'`.

**Note**: it should only be specify to use either one, but not both. Use flag or in the definition file to specify which existing sandbox to clone.

**Step to Reproduce**:

`sbx1` and `sbx2` created off production
`sbx3` which is a clone of `sbx1`
run `sf org refresh sandbox --name sbx3 --target-org prodOrgAlias --definition-file dev.json`

Inside `dev.json` file:
```
{
    "sourceSandboxName": "sbx2"
}
```
@W-18237102@

see more details: https://github.com/forcedotcom/cli/issues/3262

### DIFF
--- a/src/commands/org/refresh/sandbox.ts
+++ b/src/commands/org/refresh/sandbox.ts
@@ -190,6 +190,7 @@ export default class RefreshSandbox extends SandboxCommandBase<SandboxCommandRes
 
     let apexId: string | undefined;
     let groupId: string | undefined;
+    let srcId: string | undefined;
 
     if (defFileContent.ApexClassName) {
       apexId = await requestFunctions.getApexClassIdByName(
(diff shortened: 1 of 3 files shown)


src/commands/org/refresh/sandbox.ts
```diff
@@ -69,6 +69,18 @@ export default class RefreshSandbox extends SandboxCommandBase<SandboxCommandRes
...
+      summary: messages.getMessage('flags.source-sandbox-name.summary'),
+      description: messages.getMessage('flags.source-sandbox-name.description'),
+      exclusive: ['source-id'],
+    }),
+    'source-id': Flags.salesforceId({
+      summary: messages.getMessage('flags.source-id.summary'),
+      description: messages.getMessage('flags.source-id.description'),
+      exclusive: ['source-sandbox-name'],
+      length: 'both',
+      char: undefined,
```
WillieRuemmele, line 82: no need to specify `char:undefined`

WillieRuemmele (changes requested):

WillieRuemmele:
### QA Notes
---
✅ : `--source-sandbox-name` and name in def file => error with message
✅ `--source-id` and id in def => error
✅ id in def and `--suorce-id` => error
✅ name in def and `--source-id` => error

WillieRuemmele (approved):

PR #1443, opened 2025-06-11
feat: add features when cloning sandbox - W-18539941

soridalac:
### What does this PR do?
Resolve issue when cloning sandbox that has feature with upgrade ['data storage'].
@W-18539941@
Query in the lib: https://github.com/forcedotcom/sfdx-core/pull/1205

GH issue link: https://github.com/forcedotcom/cli/issues/3293


src/commands/org/create/sandbox.ts
```diff
@@ -175,6 +175,22 @@ export default class CreateSandbox extends SandboxCommandBase<SandboxCommandResp
...
+      if (sourceFeatures) {
+        (sandboxReq as SandboxRequest & { Features?: string }).Features = sourceFeatures;
+      }
+    } else if (srcId) {
+      const sourceFeatures = await requestFunctions.getFeature(this.flags['target-org'].getConnection(), srcId);
+      if (sourceFeatures) {
+        (sandboxReq as SandboxRequest & { Features?: string }).Features = sourceFeatures;
+      }
+    }
+
```
shetzel, line 193: 1. We should update the types for `SandboxRequest`, `SandboxInfo`, and `SandboxProcess` in sfdx-core repo so we don't have to do the intersection type here and other places.
2. There's duplicate code in the conditionals that could be simplified.  Maybe something like, `const nameOrId = srcSandboxName ?? srcId;` then if `nameOrId` is defined call getFeature()

src/shared/sandboxRequest.ts
```diff
@@ -78,7 +78,7 @@ export async function createSandboxRequest(
   definitionFile: string | undefined,
   logger?: Logger | undefined,
   properties?: Record<string, string | undefined>
-): Promise<{ sandboxReq: SandboxRequest; srcSandboxName?: string; srcId?: string }> {
+): Promise<{ sandboxReq: SandboxRequest; srcSandboxName?: string; srcId?: string; Features?: string }> {
```
shetzel, line 81: Features should be an optional prop of SandboxRequest

src/shared/sandboxRequest.ts
```diff
@@ -88,7 +88,10 @@ export async function createSandboxRequest(
 
   const capitalizedVarArgs = properties ? lowerToUpper(properties) : {};
   // varargs override file input
-  const sandboxReqWithName: SandboxRequest & { SourceSandboxName?: string; SourceId?: string } = {
+  const sandboxReqWithName: (SandboxRequest & { Features?: string }) & {
+    SourceSandboxName?: string;
+    SourceId?: string;
+  } = {
```
shetzel, line 94: `SourceId` is already a prop of SandboxRequest.  Is `SourceSandboxName` the same as `SandboxName`?  We should try to keep the types defined in 1 place and avoid intersections unless we have to.

src/shared/sandboxRequest.ts
```diff
@@ -143,11 +149,68 @@ export async function getSrcIdByName(conn: Connection, sandboxName: string): Pro
...
+    ? `SELECT Features FROM SandboxProcess WHERE Id = '${identifier}'`
+    : `SELECT Features FROM SandboxProcess WHERE SandboxName = '${identifier}'`;
+
+  try {
+    const result = await conn.singleRecordQuery<{ Features: string }>(query, { tooling: true });
+    return result.Features;
+  } catch (err) {
+    throw cloneMessages.createError('error.featuresQueryFailed', [identifier], [], err as Error);
+  }
+}
```
shetzel, line 167: The Org class already has various `querySandboxProcess()` methods so I'd rather use that code.

src/shared/sandboxRequest.ts
```diff
@@ -143,11 +149,68 @@ export async function getSrcIdByName(conn: Connection, sandboxName: string): Pro
...
+    ? `SELECT Features FROM SandboxInfo WHERE Id = '${identifier}'`
+    : `SELECT Features FROM SandboxInfo WHERE SandboxName = '${identifier}'`;
+
+  try {
+    const result = await conn.singleRecordQuery<{ Features: string }>(query, { tooling: true });
+    return result.Features;
+  } catch (err) {
+    throw cloneMessages.createError('error.featuresQueryFailed', [identifier], [], err as Error);
+  }
+}
```
shetzel, line 184: If there's no query for SandboxInfo in the `Org` class I would add this code there.

shetzel (changes requested):
A lot of code should be moved to sfdx-core.  We'll also need tests to ensure it works as intended and doesn't regress.

src/commands/org/create/sandbox.ts
```diff
@@ -175,6 +174,14 @@ export default class CreateSandbox extends SandboxCommandBase<SandboxCommandResp
       delete sandboxReq.ActivationUserGroupName;
     }
 
+    // Get source sandbox features if cloning
+    if (srcSandboxName ?? srcId) {
+      const sourceFeatures = await this.flags['target-org'].getSandboxFeatures(srcSandboxName ?? srcId);
```
shetzel, line 179: I would move the `getSandboxFeatures` function that is in sfdx-core to this file as a helper.  It would first call `querySandboxInfo()` to look for features.  If found on SandboxInfo then we'd stop and use it.  If not, we'd then `querySandboxProcess()` for features.  If found use it and if not there are no features.

shetzel (approved):

PR #1484, opened 2025-08-05
Update org open for single-use- W-19126474

soridalac:
### What does this PR do?
remove `sf org open` deprecations and single-use frontdoorUrl
remove support for `SF_SINGLE_USE_ORG_OPEN_URL` env var
@W-19126474@

PR about upcoming change: https://github.com/forcedotcom/cli/issues/3249

### DIFF
--- a/src/shared/orgOpenCommandBase.ts
+++ b/src/shared/orgOpenCommandBase.ts
@@ -5,16 +5,12 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import path from 'node:path';
-import fs from 'node:fs';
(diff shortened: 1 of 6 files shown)


src/shared/orgOpenCommandBase.ts
```diff
@@ -77,37 +73,6 @@ export abstract class OrgOpenCommandBase<T> extends SfCommand<T> {
     }
 
     if (this.jsonEnabled()) {
```
EstebanRomero84, line 75: @soridalac The if statement is no longer needed, leave just the call to `utils.openIrl() `method.

EstebanRomero84 (approved):
Looks good.

PR #1667, opened 2026-05-12
@W-22438536 Auth commands

iowillhoit:
### What does this PR do?
- Adds three new sf org auth show-* commands to safely reveal sensitive org credentials:
  - `sf org auth show-access-token`
  - `sf org auth show-sfdx-auth-url`
  - `sf org auth show-user-password`

These commands include confirmation prompts (with 30s timeout) to reduce accidental credential exposure in agent/CI logs. The `--no-prompt` and `--json` flags bypass the prompt but still emit a warning.

[@W-22438536@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-22438536)

### DIFF
--- a/messages/org.auth.show-user-password.md
+++ b/messages/org.auth.show-user-password.md
@@ -22,7 +22,7 @@ This command exposes a user password. While a password alone is not sufficient t
 
 # error.noPassword
 
-No password found for "%s". A password is only available if one was previously generated, such as by running "sf org generate password" or "sf org create user".
+No password found for "%s". A password is available only if one was previously generated locally in your DX project by running either "sf org generate password" or "sf org create user". If you generated a password in your org using Setup, it's not available to this command.
 
 # examples
 
--- a/messages/org.auth.show-user-password.md
+++ b/messages/org.auth.show-user-password.md
@@ -4,7 +4,7 @@ Show the stored password for an org's user.
 
 # description
 
-Passwords are only available for orgs where a password was previously generated using "sf org generate password". Because passwords are sensitive credentials, this command prompts for confirmation before revealing it. Skip confirmation by specifying either the --no-prompt or --json flag.
+Passwords are only available for orgs where a password was previously generated, such as by running "sf org generate password" or "sf org create user". Because passwords are sensitive credentials, this command prompts for confirmation before revealing it. Skip confirmation by specifying either the --no-prompt or --json flag.
 
 # flags.no-prompt.summary
 
@@ -20,7 +20,7 @@ This command exposes a user password. While a password alone is not sufficient t
 
 # error.noPassword
 
-No password found for "%s". A password is only available if one was previously generated using "sf org generate password".
+No password found for "%s". A password is only available if one was previously generated, such as by running "sf org generate password" or "sf org create user".
 
 # examples
 
--- a/messages/org.auth.show-user-password.md
+++ b/messages/org.auth.show-user-password.md
@@ -1,0 +1,41 @@
+# summary
+
+Show the stored password for an org's user.
+
+# description
+
+Passwords are only available for orgs where a password was previously generated using "sf org generate password". Because passwords are sensitive credentials, this command prompts for confirmation before revealing it. Skip confirmation by specifying either the --no-prompt or --json flag.
+
+# flags.no-prompt.summary
+
+Skip the security warning and reveal the password without confirmation.
+
+# prompt.show-user-password
+
+You're about to reveal the password for "%s". Do you want to continue?
+
+# warning.show-user-password
+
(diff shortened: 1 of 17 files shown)


src/commands/org/auth/show-access-token.ts
```diff
@@ -0,0 +1,83 @@
...
+    }),
+  };
+
+  private org!: Org;
+
+  public async run(): Promise<OrgAuthShowAccessTokenResult> {
+    const { flags } = await this.parse(OrgAuthShowAccessToken);
+
+    this.org = flags['target-org'];
+    this.org.getConnection(flags['api-version']);
```
cristiand391: `getConnection` arg is optional so we could remove `--api-version` from this command. It will use the API version set in the auth file

cristiand391 (approved):

messages/org.auth.show-user-password.md
```diff
@@ -0,0 +1,41 @@
...
+
+Skip the security warning and reveal the password without confirmation.
+
+# prompt.show-user-password
+
+You're about to reveal the password for "%s". Do you want to continue?
+
+# warning.show-user-password
+
+This command exposes a user password. While a password alone is not sufficient to gain access to an org (additional factors like a security token or an enabled OAuth username-password flow are required), treat it as a sensitive credential and avoid sharing or logging it.
```
jshackell-sfdc:
```suggestion
This command exposes a user password. While a password alone isn't sufficient to gain access to an org (additional factors like a security token or an enabled OAuth username-password flow are required), treat it as a sensitive credential and avoid sharing or logging it.
```

messages/org.auth.show-user-password.md
```diff
@@ -0,0 +1,41 @@
+# summary
+
+Show the stored password for an org's user.
+
+# description
+
+Passwords are only available for orgs where a password was previously generated, such as by running "sf org generate password" or "sf org create user". Because passwords are sensitive credentials, this command prompts for confirmation before revealing it. Skip confirmation by specifying either the --no-prompt or --json flag.
```
jshackell-sfdc: I DM'd you about this, but I'll ask here too.  Is the idea that this CLI command can be used to show only the passwords that were generated in the DX project using a CLI command, such as `org generate password` or `org create user`?  In other words, if I create a user/password in Setup in the org, that password WON'T be showable with this command?  If so, how about this description:

---start---
This command shows only passwords that were generated locally in your DX project with either the `org generate password` or `org create user` CLI command.  If you generated a password for a user in Setup in your org, you can't show it with this command.  

Because passwords are sensitive credentials, this command prompts for confirmation before revealing it. Skip confirmation by specifying either the --no-prompt or --json flag.

---end---

messages/org.auth.show-user-password.md
```diff
@@ -0,0 +1,41 @@
...
+
+You're about to reveal the password for "%s". Do you want to continue?
+
+# warning.show-user-password
+
+This command exposes a user password. While a password alone is not sufficient to gain access to an org (additional factors like a security token or an enabled OAuth username-password flow are required), treat it as a sensitive credential and avoid sharing or logging it.
+
+# error.noPassword
+
+No password found for "%s". A password is only available if one was previously generated, such as by running "sf org generate password" or "sf org create user".
```
jshackell-sfdc: similar to previous comment.  If what I suspect is true, let's change this error to:

No password found for "%s". A password is available only if one was previously generated locally in your DX project by running either `sf org generate password` or `sf org create user`. If you generated a password in your org using Setup, it's not available to this command.

jshackell-sfdc (commented):
@iowillhoit , see my questions, and possible suggestions.

cristiand391 (approved):

messages/org.auth.show-user-password.md
iowillhoit: That's correct, if you created a user/pw in Setup and then authed to it locally, the pw would not be stored locally.

iowillhoit:
```suggestion
This command shows only passwords that were generated locally in your DX project with either the "org generate password" or "org create user" CLI command. If you generated a password for a user in Setup in your org, you can't show it with this command.

Because passwords are sensitive credentials, this command prompts for confirmation before revealing it. Skip confirmation by specifying either the --no-prompt or --json flag.
```

messages/org.auth.show-user-password.md
iowillhoit:
```suggestion
No password found for "%s". A password is available only if one was previously generated locally in your DX project by running either "sf org generate password" or "sf org create user". If you generated a password in your org using Setup, it's not available to this command.
```

jshackell-sfdc (approved):
lgtm

PR #1673, opened 2026-05-18
@W-22439379: remove secrets

iowillhoit:
### What does this PR do?
- Redacts `accessToken` and `password` from `sf org list --json` output
- Redacts `accessToken` from `sf org create scratch --json` and `sf org resume scratch --json` output
- Adds `SF_TEMP_SHOW_SECRETS=true` env var as a temporary workaround for CI pipelines that depend on these values
- Centralizes redacted messages in `messages/secrets-redacted.md` for reuse across commands

## Related
- https://github.com/salesforcecli/plugin-org/pull/1667 (new `sf org auth show-*` commands)

## Test plan

- [ ] `sf org list --json` no longer shows real access tokens
- [ ] `sf org create scratch --json` no longer shows access token
- [ ] `sf org resume scratch --json` same as above
- [ ] Setting `SF_TEMP_SHOW_SECRETS=true` restores original behavior with deprecation warning
- [ ] Unit tests pass: `yarn test`

[@W-22439379@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-22439379)

### DIFF
--- a/src/commands/org/list.ts
+++ b/src/commands/org/list.ts
@@ -16,7 +16,17 @@
 
 import { EOL } from 'node:os';
 import { Flags, loglevel, SfCommand } from '@salesforce/sf-plugins-core';
-import { AuthInfo, ConfigAggregator, ConfigInfo, Connection, Org, SfError, Messages, Logger } from '@salesforce/core';
+import {
+  AuthInfo,
+  ConfigAggregator,
+  ConfigInfo,
+  Connection,
+  envVars,
+  Org,
+  SfError,
+  Messages,
+  Logger,
+} from '@salesforce/core';
 import { Interfaces } from '@oclif/core';
 import ansis, { type Ansis } from 'ansis';
 import { OrgListUtil, identifyActiveOrgByStatus } from '../../shared/orgListUtil.js';
@@ -25,6 +35,7 @@ import { ExtendedAuthFields, FullyPopulatedScratchOrgFields } from '../../shared
 
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/plugin-org', 'list');
+const secretsMessages = Messages.loadMessages('@salesforce/plugin-org', 'secrets-redacted');
 
 export const defaultOrgEmoji = '🍁';
 export const defaultHubEmoji = '🌳';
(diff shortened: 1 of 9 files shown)


src/commands/org/list.ts
```diff
@@ -126,6 +137,17 @@ Legend:  ${defaultHubEmoji}=Default DevHub, ${defaultOrgEmoji}=Default Org ${
...
     );
 
+    // TODO: Remove after env var workaround is removed
+    const showSecretsEnvVarIsSet = envVars.getBoolean('SF_TEMP_SHOW_SECRETS', false);
+
+    if (this.jsonEnabled()) {
+      if (showSecretsEnvVarIsSet) {
+        this.warn(secretsMessages.getMessage('temp.envVarIsSet', ['sf org list']));
+      } else {
+        this.warn(secretsMessages.getMessage('temp.envVarWorkaround', ['sf org list']));
```
cristiand391, line 147: nit:
could it print the warning all the time instead of only for `--json`?
it's an old command so most `--json` calls are likely being piped to `jq` or something else, users updating to latest and running `sf org list` in their local machines could notice the message earlier.

cristiand391 (approved):

src/commands/org/list.ts
iowillhoit: The HRO doesn't show tokens though. I think it would be confusing to have a warning about redacted messages when none are displayed.
My thought was that if someone is piping an AT from `org list`, it will fail because the `accessToken` will now be the redacted message. They will then likely see the warnings since warnings are sent to `stderr`

cristiand391: ahh right, forgot it's actually a breaking change. Nvm 👍🏼

cristiand391:
**QA notes**:

### `sf org display`

✅ access token redacted in human-readable and json output
✅ with `--verbose`, sfdx auth url also redacted
✅ auth'd to an org via `sf org login web --client-id` and passed a client secret. It never prints it as decrypted (also checked on latest)
✅ `SF_TEMP_SHOW_SECRETS=true` shows all redacted values with warning in json

### `sf org create scratch` / `sf org resume scratch`

✅ `--async --json` doesn't include anything sensitive from devhub (client id is considered public)
✅ `org resume scratch --json` doesn't include AuthCode, redacts scratch org access token
✅ sync create scratch doesn't include AuthCode, redacts scratch org access token
🟡 `sf org resume scratch --job-id 2SRKY00000XpoqM4AR --json ` still includes `.result.authFields.privateKey`


### `sf org list`
✅ access token redacted in json output
✅ `SF_TEMP_SHOW_SECRETS=true` shows all redacted values with warning in json

PR #1748, opened 2026-08-14
fix: bypass default-browser PATH hijack on Windows for --private @W-23807283@

WillieRuemmele:
## Summary
- On Windows, `sf org open --private` calls the `default-browser` npm package which invokes bare `reg` (no fully-qualified path), allowing a malicious repo-local `reg.exe` to execute via PATH hijack
- This PR bypasses `default-browser` entirely on Windows by detecting the default browser ourselves using a fully-qualified `%SYSTEMROOT%\System32\reg.exe` path, then passing an explicit app name + incognito flag to `open()` (which skips `default-browser` for any concrete app name)
- Mirrors the safe pattern from `wsl-utils`'s `powerShellPath()` which already qualifies system binary paths
- Upstream fix contributed and merged: https://github.com/sindresorhus/default-browser/pull/23 (released in `default-browser@5.5.1`)
- Bumped `open` to `^11.0.1` to pull in the upstream fix via lockfile; workaround kept as defense-in-depth

## Work Item
@W-23807283@: SF CLI - Project-local reg.exe leads to RCE during private org opening

## Test plan
- [x] Unit tests for `getWindowsPrivateBrowserApp()` covering Chrome, Firefox, Edge, Brave, hyphen-suffixed ProgIds, unsupported browsers, and unparseable output
- [x] Unit test verifying the fully-qualified path uses SYSTEMROOT env var
- [x] Windows-only NUT that verifies a poisoned PATH does not result in execution of a local `reg.exe`
- [ ] Manual verification on Windows with a repo containing `node_modules/.bin/reg.cmd`

### DIFF
--- a/test/nut/openPrivatePathHijack.nut.ts
+++ b/test/nut/openPrivatePathHijack.nut.ts
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2026, Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { platform } from 'node:os';
+import fs from 'node:fs';
+import path from 'node:path';
+import { tmpdir } from 'node:os';
+import { expect } from 'chai';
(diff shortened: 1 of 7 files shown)


test/nut/openPrivatePathHijack.nut.ts
```diff
@@ -0,0 +1,76 @@
...
+    expect(capturedCommand).to.not.equal('reg.exe');
+  });
+
+  it('a project-local reg.exe in PATH does not execute during browser detection', async () => {
+    const poisonDir = path.join(tmpdir(), `hijack-test-${process.pid}`);
+    const poisonBin = path.join(poisonDir, 'node_modules', '.bin');
+    fs.mkdirSync(poisonBin, { recursive: true });
+
+    fs.writeFileSync(
+      path.join(poisonBin, 'reg.cmd'),
```
k80bowman: This should be a reg.exe file, right? The bug will not trigger on a reg.cmd file.

k80bowman (approved):
Approving this, though I do have a comment on one of the tests.

PR #1775, opened 2026-09-10
fix: bump open package to latest-W-24154138

soridalac:
### Summary

  - Fixes race condition where sf org open exits before the browser-opener child process finishes launching the browser, causing intermittent
    failures — particularly on Windows and in VS Code's integrated terminal
  - Awaits the child process exit instead of fire-and-forget, with a guard for the case where the process has already exited before the handler is
    registered

### Test plan

  - [ ] sf org open opens browser reliably on Windows (standalone terminal and VS Code integrated terminal)
  - [ ] sf org open continues to work on macOS and Linux with no added delay
  - [ ] sf org open --url-only is unaffected (skips browser launch)
  - [ ] sf org open --browser firefox works
  - [ ] sf org open --private works
  - [x] Unit tests pass
@W-24154138@
cli related issue: https://github.com/forcedotcom/cli/issues/3646

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ EXAMPLES
     $ sf org auth show-access-token --target-org my-org --json
 ```
 
-_See code: [src/commands/org/auth/show-access-token.ts](https://github.com/salesforcecli/plugin-org/blob/6.0.13/src/commands/org/auth/show-access-token.ts)_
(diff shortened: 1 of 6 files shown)


WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
✅ : verified by customers https://github.com/forcedotcom/cli/issues/3646
✅ : bug acknowledged, and fixed in `open` package

